PR #61, opened 2018-11-25
Qt 5.11 docker

zaksabeast:
This pull request updates Appveyor's docker usage to use Qt 5.11.


Admiral-Fish:
For reference this is due to 7d96991 using a Qt 5.11 feature and the old system using Qt 5.9 for linux.

PR #117, opened 2020-08-28
fixed potential use of un-initialized variables

Katana-Steel:


Admiral-Fish:
This looks to be the output of an automated tool. As far as I can tell all these cases the values are given a valid value before use.

Katana-Steel:
Yeah the automated tool is GCC 9.2 or newer

PR #126, opened 2020-10-21
Added Advances research in Researcher window

Real96:
* Added Advances research in Researcher window (useful for research purpose)

image


Admiral-Fish:
I'm not sure if I see the point of this. Advances can be controlled by initial and max plus its sequential so it's not like searching is extremely helpful if you already know it exists.

Issue #65, opened 2019-02-11
[Gen 3] RSE (ONLY) Safari Zone Fishing and maybe Rock Smash


RichardPaulAstley:
![image](https://user-images.githubusercontent.com/25870563/52600494-1c68a600-2e5c-11e9-8a40-c3d8396f9e84.png)
^the screen of the issue i've stated in the discord

Admiral-Fish:
Dumping this link for later reference
https://github.com/pret/pokeruby/blob/master/src/wild_encounter.c#L3951

RichardPaulAstley:
I'm gonna ad the Rock Smash issue i'm getting in the Safari Zone.
It could mess up like it does for Fishing, no way to get something working there.

As reminder, in RSE, the wild function is different in the Safari Zone, and so should be applied to every case (Fishing, Rock Smash)

Making some of them broken

Real96:
Rock Smash Fix
Rock smash has to be fixed in both RS ane E, inside and outside the Safari Zone.
To get correct rock smash spreads i used researcher + safari zone encounter type. The battle triggering prng is at frame 45725, the spread+pid are generated from the next prng, so 45726. The PID/spread i got matches perfectly Safari encounter type research as you can see in this screen (PID/spread are at frame 45726).

![1](https://user-images.githubusercontent.com/20956021/80203765-fa55e080-8627-11ea-8182-5c1bca4419fa.png)

Comparing rock smash encounter type results to it, frame is wrong (should be +1, as you can see the pid is the same of safari encounter type research but frame is -1) and it generates the spread from the wrong prng seed. Atm it is using the prng which trigger the battle but instead it should use the next one.

![2](https://user-images.githubusercontent.com/20956021/80204358-00988c80-8629-11ea-831e-650f90be156a.png)

This happens also outside safari zone, you can apply the same thing comparing grass / rock smash.

Real96:
Fishing fix
Fishing has to be fixed for both RS and E inside Safari zone only (it is ok outside).
PID and spread are ok, the only problem is that level and encounter slots are in the previous frame as you can see in this screen:

![3](https://user-images.githubusercontent.com/20956021/80207240-19577100-862e-11ea-8331-f2ce42e8d327.png)

This happens with all 3 rods

RichardPaulAstley:
You can add fishing to the list

Real96:
Fixed in #127

PR #127, opened 2020-10-21
[Gen3] Fix Rock Smash and Fishing generation in Safari Zone for RSE (#65)

Real96:
* Fixed Rock Smash generation in Safari Zone for RSE (#65) - Tested and now working both inside and outside Safari Zone
* Fixed Fishing generation in Safari Zone for RSE (#65) - Tested and now working both inside and outside Safari Zone


Admiral-Fish (changes requested):
Looking at the optimized assembly it is inlining the call to isSafariZoneEncounter but will call encounterArea.getLocation() several times when it really only needs to call it once.

The safari zone function should be moved to EncounterArea3 to reduce the number of function calls it does.

Also does WildSearcher3 need to be modified to match these changes?

Real96:
> Looking at the optimized assembly it is inlining the call to isSafariZoneEncounter but will call encounterArea.getLocation() several times when it really only needs to call it once.
> 
> The safari zone function should be moved to EncounterArea3 to reduce the number of function calls it does.
> 
> Also does WildSearcher3 need to be modified to match these changes?

It should be ok now!
Researcher seems to work correctly, but it display wrong Lv and wrong Leader ability

Real96:
Everything fixed except RSE Safari Zone Rock Smash searcher

Real96:
Fixed everything! Tested and perfectly working. If you don't see anything that need to be changed in the code, can merge!

Source/Core/Gen3/Searchers/WildSearcher3.cpp
```diff
@@ -22,9 +22,10 @@
 #include <Core/RNG/LCRNG.hpp>
 #include <Core/Util/EncounterSlot.hpp>
 
-WildSearcher3::WildSearcher3(u16 tid, u16 sid, u8 genderRatio, Method method, const StateFilter &filter) :
+WildSearcher3::WildSearcher3(u16 tid, u16 sid, u8 genderRatio, Method method, const StateFilter &filter, bool isRSEVersion) :
```
Admiral-Fish, line 25: Why are of any the changes in this file necessary? I don't account for method so why would whether or not it is safari zone matter. Some of the methods vary in rng calls as well.

Admiral-Fish (commented):
A few general questions since I don't know how to proceed with this.

1. It has been so long since I have looked at these issues I barely understand why changes are necessary. Could you explain if possible or add some comments about why the extra RNG calls are necessary?

2. In the past I had considered adding new Encounter types. SafariGrass, SafariSurfing, etc (you get the idea). What are your thoughts on this? The immediate thing is would remove all the if checks, but would add some duplication since the encounters are fairly similar.

Source/Core/Gen3/Searchers/WildSearcher3.cpp
Real96: Fixed!

Real96: Safari zone has got a different generation, so they're needed to match Generator results in case you are searching a spread for RSE safari zone

Real96:
> 1. It has been so long since I have looked at these issues I barely understand why changes are necessary. Could you explain if possible or add some comments about why the extra RNG calls are necessary?

These changes are necessary cause inside RSE Safari Zone, Fishing and Rock Smash generation are a bit different than outside (you can check this in #65). All these changes i made has been tested by me and Rick, so they fixed the generation perfectly.

I commented all RNG calls!

>
> 2. In the past I had considered adding new Encounter types. SafariGrass, SafariSurfing, etc (you get the idea). What are your thoughts on this? The immediate thing is would remove all the if checks, but would add some duplication since the encounters are fairly similar.

I think they are not necessary, cause it will be a RSE only thing, while FRLG won't need this. I think it would be better to leave the if checks rather than adding them

Issue #123, opened 2020-10-18
Advances column of Researcher window should start from 1 instead of 0


Real96:
Advances in researcher should start from one instead of 0.
If you use for example LCRNG and your starting seed is 0, it should be Advances 0 and 6073 should be advance 1: you're are at seed 0 (so 0 advances) and you move to 6073 (1 advance).

This is an evidence example of wrong advances in researcher:

![image](https://user-images.githubusercontent.com/20956021/96366270-e0183d80-1146-11eb-822e-409166651f4c.png)

As you can see Advances in Researcher is -1.
It is pretty useful to have it starting from 1 for researching purpose.

Real96:
Pull request #125 fix this making the advances still starting from 0 but showing the inserted initial seed as the first seed.

![image](https://user-images.githubusercontent.com/20956021/96654569-ede2e400-133b-11eb-93ba-bfa48493e25f.png)

PR #128, opened 2020-10-21
Make inserted starting seed the first seed in Researcher output (#123)

Real96:
* Inserted starting seed will be the first seed in Researcher output

image

### DIFF
--- a/Forms/Util/Researcher.cpp
+++ b/Forms/Util/Researcher.cpp
@@ -240,21 +240,27 @@ void Researcher::generate()
         {
         case 0:
             rngStates = getStates(PokeRNG(seed), initialAdvances, maxAdvances);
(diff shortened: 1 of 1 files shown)


Admiral-Fish:
I'm not a big fan of this approach. I don't think everything needs the initial seed included in the output such as the MT/SFMT/TinyMT. I think it might be better if each one that needs the initial seed added does it manually to the vector so that the for loop doesnt need to be modified

Real96:
> I'm not a big fan of this approach. I don't think everything needs the initial seed included in the output such as the MT/SFMT/TinyMT. I think it might be better if each one that needs the initial seed added does it manually to the vector so that the for loop doesnt need to be modified

Fixed in the way you told me on discord (`rngStates.prepend(seed);`)

PR #136, opened 2020-12-21
Add support for ARM devices

DarkKirb:
This commit changes the SSE/AVX intrinsics code to use a slightly more
cross platform version using defines, making porting to other platforms
easier.

It also implements these macros for ARM NEON, the SIMD extension found
on ARM processors. While that instruction set is also used on M1-Macs, I
do not know if there is work necessary elsewhere before it runs on these
devices. It works on my (dated) ARM laptop however.

NB: Currently it removes the -msse4.2 command line argument from the project file, it probably needs to be replaced with a conditional expression that checks the target architecture, but i have zero experience with qmake.


Admiral-Fish:
I am in the process of converting the build system to cmake. The build flags can be handled on that (not sure if you have cmake experience). Also, does ARM require any compiler flags for SSE/AVX2 instructions?

Does ARM have AVX2 instructions? I only noticed you using SSE instructions.

I am also not sure about the M1-Macs, would have to check what versions of Qt support it. But the rest of the program should be fine to compile for ARM.

DarkKirb:
ARM has neither SSE nor AVX instruction sets, as they are only supported on x86(_64). Arm NEON does not support 256 bit or larger vectors and my ARM laptop does not support the successor to NEON, SVE (which supports almost arbitrary vector sizes, the native vector size is cpu-dependent though, so it might not be any faster on lower-end CPUs)

On arm64, NEON is available by default, on 32-bit arm you need to pass `-mfpu=neon`.

Admiral-Fish:
Can you add the necessary compile flags? Then I can merge this.

Admiral-Fish:
Windows and Linux builds are failing. Seems like what you did only works for Mac which didn't seem to care as much about the compiler flags.

cmake/GetTargetArch.cmake
```diff
@@ -0,0 +1,16 @@
...
+    ERROR_QUIET
+  )
+  if (RESULT)
+    message(FATAL_ERROR "Failed  to determine target architecture: ${RESULT}")
+  else()
+    string(REGEX MATCH "([^-]+).*" ARCH_MATCH ${ARCH})
+    set(ARCH ${CMAKE_MATCH_1})
+  endif()
+  message(STATUS "Target arch: ${ARCH}")
+  set(FLAG ${ARCH} PARENT_SCOPE)
```
Admiral-Fish, line 15: ${FLAG}

cmake/GetTargetArch.cmake
```diff
@@ -0,0 +1,16 @@
+function(get_target_arch FLAG)
+  execute_process(
+    COMMAND ${CMAKE_CXX_COMPILER} ${CMAKE_CXX_FLAGS} -dumpmachine
```
Admiral-Fish, line 3: This just doesn't work on Windows.

Admiral-Fish (changes requested):
Still not working. You can view appveyor to see the status of builds if you are unable to test locally.

cmake/GetTargetArch.cmake
DarkKirb: thanks

cmake/GetTargetArch.cmake
DarkKirb: I didn't mean to run that code on windows, i moved it down to after the if(UNIX)

PR #148, opened 2021-03-17
Use ISO 8601 format for dates

KPJoshi:
In the USA, dates are often written as **month/day/year**. In many other countries, they are written as **day/month/year**. I believe that the second format is used for PAL region consoles. I propose using the ISO 8601 standard format for dates. This prints dates as **year-month-day**. For US region consoles, this will still be easy to understand since the year is the only 4 digit number and the order of month and day is preserved. For PAL region users, I believe my proposed format will be more understandable than the current default US format.

There might be 2 better alternatives:
- Representing the month as a 3 letter abbreviation. However in DS/3DS you enter the month as a number, so this will be slightly inconvenient, if less unambiguous.
- Adding a setting to decide the output format. But this would be out of the way in the dropdown menu.

Links:
- [ISO 8601 on Wikipedia](https://en.wikipedia.org/wiki/ISO_8601)
- [Date format by country on Wikipedia](https://en.wikipedia.org/wiki/Date_format_by_country)

### DIFF
--- a/Source/Core/Util/DateTime.cpp
+++ b/Source/Core/Util/DateTime.cpp
@@ -107,7 +107,7 @@ std::string Date::toString() const
     std::string d = std::to_string(parts[2]);
     d.insert(d.begin(), 2 - d.size(), '0');
 
-    return m + "/" + d + "/" + y;
+    return y + "-" + m + "-" + d;
 }
 
 int Time::addSeconds(int seconds)
--- a/Source/Forms/Controls/DateEdit.cpp
+++ b/Source/Forms/Controls/DateEdit.cpp
@@ -21,6 +21,7 @@
 
 DateEdit::DateEdit(QWidget *parent) : QDateEdit(parent)
 {
+    setDisplayFormat("yyyy-MM-dd");
 }
 
 Date DateEdit::getDate() const
--- a/Source/Forms/Controls/DateTimeEdit.cpp
+++ b/Source/Forms/Controls/DateTimeEdit.cpp
@@ -21,7 +21,7 @@
 
 DateTimeEdit::DateTimeEdit(QWidget *parent) : QDateTimeEdit(parent)
 {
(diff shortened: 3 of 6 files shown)


Admiral-Fish:
This isn't the only place dates are used. A number of forms have date controls, shouldn't those be changed as well?

KPJoshi:
> This isn't the only place dates are used. A number of forms have date controls, shouldn't those be changed as well?

This is the function within the date/time file that is responsible for converting it to a string. So any place that displays the date uses this function. I checked in multiple locations in the compiled binary. Also I could not find any place where you enter the date yourself (except the year in some places)

Admiral-Fish:
There are several places in the UI that have selection for date and time. The constructors for DateEdit and DateTimeEdit should probably have a call to setDisplayFormat to follow ISO 8601.

KPJoshi:
> There are several places in the UI that have selection for date and time. The constructors for DateEdit and DateTimeEdit should probably have a call to setDisplayFormat to follow ISO 8601.

All the places where you can enter date or date+time _already_ use my proposed year-month-day format! Check, for example, the TID/SID finder for any gen. It seems that this is the default for QT when it comes to date and time inputs. I do not think adding any calls to setDisplayFormat will be required.

Admiral-Fish:
I think it defaults to the system locale. It shows as M/D/Y for me.

KPJoshi:
I added the formats to the constructors of DateEdit and DateTimeEdit. I also removed several inconsistencies:
- The Gen 4 TID/SID finder did not use the custom DateTimeEdit class but QT's original class. I changed it to use the custom class (otherwise the format would not be set)
- Some places had specified a conflicting date/time format that overrode the one provided in the constructor. I removed those conflicts.

Sorry for the multiple commits fixing the same problem. Please squash when you merge.

Admiral-Fish:
Will merge sometime this weekend. Been busy this week.

PR #155, opened 2021-04-15
Fixed gen5 IDs rng advances being -1

Real96:
Necessary because game does 1 advancement when you enter the name insertion screen for the first time. ID and SID are correct, advances instead are -1 compared to in game last screen advances. Tested many times and now perfectly working.


Admiral-Fish:
Can you add a comment above the for loop explaining why it starts from 1 instead of 0. Then I will merge

Real96:
> Can you add a comment above the for loop explaining why it starts from 1 instead of 0. Then I will merge

Done!

PR #156, opened 2021-04-15
Added Key Advances column Gen5 DreamRadar

Real96:
Advances in DreamRadar rng now matches perfectly in game advances. Tested many times and now perfectly working, also the researcher.

### DIFF
--- a/Source/Core/Gen5/Generators/DreamRadarGenerator.cpp
+++ b/Source/Core/Gen5/Generators/DreamRadarGenerator.cpp
@@ -47,12 +47,13 @@ DreamRadarGenerator::DreamRadarGenerator(u32 initialAdvances, u32 maxAdvances, u
     }
 }
 
-std::vector<State> DreamRadarGenerator::generate(u64 seed, bool memory)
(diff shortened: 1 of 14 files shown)


Admiral-Fish:
Why is this necessary?

Real96:
> Why is this necessary?

To match game advances. It would be better imo, first time i tested it i didn't understood what i was doing wrong. I think this would be less confusing. All the other rng just follow game advances, this is the only different one.

Real96:
In these screen there are original pokefinder at the left, my pr pokefinder and the in game advances showed by DeSmuMe. As you can see initial advances aren't 0. My pr match perfectly the game,

Real96:
Modified as agreed on discord. Both Generator and Searcher are fully working!
image

Admiral-Fish:
Can you fix merge conflicts and then make DreamRadarState header only? Then I can merge.

Real96:
> Can you fix merge conflicts and then make DreamRadarState header only? Then I can merge.

Fixed merge conflicts and made DreamRadarState header only. Pls merge before new merge conflicts appear t.t

PR #187, opened 2021-09-15
[Gen5] Fix hidden grotto generator/searcher advances -1 offset (#186)

### LINKED ISSUE #186: [Gen5] Hidden Grotto generator/searcher advances are -1
This was introduces in [this](https://github.com/SteveCookTU/PokeFinder/commit/68d93587d575a34659d59f79625445c423d1a1fc) commit and causes the Hidden Grotto generator/searcher to display advances values -1. All the other columns are perfect.

image

On the left the correct one, on the right the actual and wrong one.

Real96:
Fix #186

### DIFF
--- a/Source/Core/Gen5/Generators/HiddenGrottoGenerator.cpp
+++ b/Source/Core/Gen5/Generators/HiddenGrottoGenerator.cpp
@@ -45,7 +45,7 @@ std::vector<HiddenGrottoState> HiddenGrottoGenerator::generate(u64 seed) const
     for (u32 cnt = 0; cnt <= maxAdvances; cnt++, rng.next())
     {
         BWRNG go(rng.getSeed());
-        u32 seed = go.nextUInt(0x1FFF);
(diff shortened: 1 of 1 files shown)


Source/Core/Gen5/Generators/HiddenGrottoGenerator.cpp
```diff
@@ -45,7 +45,7 @@ std::vector<HiddenGrottoState> HiddenGrottoGenerator::generate(u64 seed) const
     for (u32 cnt = 0; cnt <= maxAdvances; cnt++, rng.next())
     {
         BWRNG go(rng.getSeed());
-        u32 seed = go.nextUInt(0x1FFF);
+        u32 seed = ((go.getSeed() >> 32) * 0x1FFF) >> 32;
```
Admiral-Fish, line 48: This is correct in-game against chatot pitches?

Real96: yes it is, i checked in game directly

PR #197, opened 2021-10-04
Added missing FRLG breeding rng method

Real96:
Code taken from [here](https://github.com/Lincoln-LM/Finder-ToolBox/commit/bf75fd531369af8a8bacaaebfc0a2d8fcde4d7eb) and [here](https://github.com/Lincoln-LM/Finder-ToolBox/commit/c129e822dde485d389db5a2cf6b1d1bf60cce6c2).

Thanks to [Lincoln-LM](https://github.com/Lincoln-LM/)!

### DIFF
--- a/Source/Forms/Gen3/Eggs3.ui
+++ b/Source/Forms/Gen3/Eggs3.ui
@@ -103,9 +103,6 @@
    </item>
    <item row="1" column="0">
     <widget class="QTabWidget" name="tabEggSelection">
-     <property name="currentIndex">
-      <number>2</number>
-     </property>
      <widget class="QWidget" name="tabEmerald">
       <attribute name="title">
        <string>Emerald</string>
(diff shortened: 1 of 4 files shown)


Source/Forms/Gen3/Eggs3.ui
```diff
@@ -103,6 +103,9 @@
    </item>
    <item row="1" column="0">
     <widget class="QTabWidget" name="tabEggSelection">
+     <property name="currentIndex">
+      <number>2</number>
+     </property>
```
Admiral-Fish, line 108: Please remove this

Admiral-Fish (changes requested):
Does RS require this as well?

Source/Forms/Gen3/Eggs3.ui
Real96: Done

Real96:
> Does RS require this as well?

No one hit this method on RS, so i assume it isn't needed. If we get this in future i'll add it in case

Issue #184, opened 2021-09-13
Add missing leads


Admiral-Fish:
Static
Magnet pull
Pressure (and friends)
Compound eyes (and friends)

Maybe level ones, probably not
pls help

RichardPaulAstley:
Everything is important tbh. Levels can give better solution for some specific case. 

These are the main ones for the levels : 

Hustle / Vital Spirit / Pressure : 50% (on the generation) to find a Pokémon with a higher level than your lead
Intimidate / Keen Eye  : 50% (on the generation) to avoid a Pokémon with a lower level than your lead.

I avoid the ones that implies Sweet Scentless stuff (tell me if you're interested) 

These are a thing since Emerald but I'd double check on that.

ShinySylveon04:
More info for Compound Eyes:

- Normal rates: 45 / 95
- Compound Eyes: 20 / 80

Admiral-Fish:
Closing as fixed via redesign branch. Gen 5 will have the appropriate leads added when implemented

PR #200, opened 2021-11-04
[Gen 4] Added Compound Eyes lead ability

Real96:
Add what requested in #184 (gen4 compound eyes only).


Source/Forms/Models/Gen4/WildModel4.cpp
```diff
@@ -64,7 +64,14 @@ QVariant WildGeneratorModel4::data(const QModelIndex &index, int role) const
...
             u8 item = state.getItem();
-            return item < 45 ? "No" : item < 95? "50%" : "5%";
+            if (state.getLead() == Lead::CompoundEyes)
+            {
+                return item < 20 ? "No" : item < 80? "60%" : "20%";
+            }
+            else
+            {
+                return item < 45 ? "No" : item < 95? "50%" : "5%";
+            }
```
Admiral-Fish, line 74: Maybe it would be better to do a Normal/Rare item instead of listing percents. What do you think?

Source/Forms/Models/Gen4/WildModel4.cpp
```diff
@@ -284,7 +293,14 @@ QVariant WildSearcherModel4::data(const QModelIndex &index, int role) const
...
```
Admiral-Fish, line 303: Same as above

Source/Forms/Models/Gen4/WildModel4.cpp
Real96: How would do this on gen5 then? You have also 1% there

Source/Core/Gen4/Searchers/WildSearcher4.cpp
```diff
@@ -378,7 +378,7 @@ std::vector<WildState> WildSearcher4::searchInitialSeeds(const std::vector<WildS
 std::vector<WildState> WildSearcher4::normalMethodJ(WildState state, u32 seed) const
 {
     std::vector<WildState> states;
-    state.setLead(Lead::None);
+    lead == Lead::CompoundEyes? state.setLead(Lead::CompoundEyes) : state.setLead(Lead::None);
```
Admiral-Fish, line 381: Probably just be better to do `state.setLead(lead)`

Source/Core/Gen4/Searchers/WildSearcher4.cpp
```diff
@@ -546,7 +546,7 @@ std::vector<WildState> WildSearcher4::normalMethodK(WildState state, u32 seed) c
     {
         if ((nextRNG % 25) == state.getNature())
         {
-            state.setLead(Lead::None);
+            lead == Lead::CompoundEyes? state.setLead(Lead::CompoundEyes) : state.setLead(Lead::None);
```
Admiral-Fish, line 549: Probably just be better to do `state.setLead(lead)`

Source/Forms/Models/Gen4/WildModel4.cpp
Admiral-Fish: Leave the numbers then. Change `"No"` to `tr("None")`

Source/Forms/Models/Gen4/WildModel4.cpp
Admiral-Fish: Leave the numbers then. Change `"No"` to `tr("None")`

Admiral-Fish (changes requested):

Source/Forms/Models/Gen4/WildModel4.cpp
Real96: done

Source/Core/Gen4/Searchers/WildSearcher4.cpp
Real96: done

Source/Core/Gen4/Searchers/WildSearcher4.cpp
Real96: done

Source/Forms/Models/Gen4/WildModel4.cpp
Real96: done

PR #202, opened 2021-11-05
[Gen 3] Fix Emerald egg generator displaying 0 advances

### LINKED ISSUE #201: Emerald Egg Generator Displaying 0 Advances
Tested on the windows appveyor build of 14eb361d9509c108bab6e0db2dd3c96ef3478880
Appears to affect all methods of emerald eggs only

To reproduce: search with any params in the Emerald Egg generator

Real96:
Fix #201

### DIFF
--- a/Source/Forms/Models/Gen3/EggModel3.cpp
+++ b/Source/Forms/Models/Gen3/EggModel3.cpp
@@ -62,7 +62,14 @@ QVariant EggModel3::data(const QModelIndex &index, int role) const
         {
         case 0:
         case 1:
-            return state.getGenerateAdvance();
+            switch (method)
+            {
+            case Method::RSBred:
+            case Method::FRLGBred:
+                return state.getGenerateAdvance();
+            default:
+                return state.getAdvances();
+            }
         case 2:
             return state.getAdvances();
         case 3:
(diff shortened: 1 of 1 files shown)


Source/Forms/Models/Gen3/EggModel3.cpp
```diff
@@ -62,7 +62,14 @@ QVariant EggModel3::data(const QModelIndex &index, int role) const
...
         case 1:
-            return state.getGenerateAdvance();
+            switch (method)
+            {
+            case Method::RSBred:
+            case Method::FRLGBred:
+                return state.getGenerateAdvance();
+            default:
+                return state.getAdvances();
+            }
```
Admiral-Fish, line 72: I am not a big fan of this approach. It would be best to modify getColumn and the QStringList holding all the strings if necessary. This way we only do the switch statement once.

Admiral-Fish (changes requested):

Source/Forms/Models/Gen3/EggModel3.cpp
Real96: > I am not a big fan of this approach. It would be best to modify getColumn and the QStringList holding all the strings if necessary. This way we only do the switch statement once.

Changed

PR #205, opened 2021-11-23
Add egg seed to egg 8

SteveCookTU:


Admiral-Fish (changes requested):
The parent class State already has `getSeed()` and `setSeed()` functions that can be reused

PR #210, opened 2021-12-02
Wild 8

Admiral-Fish:


Source/Core/Gen8/Generators/WildGenerator8.cpp
```diff
@@ -18,26 +18,181 @@
...
+                continue;
+            }
+
+            state.setLevel(encounterArea.calcLevel(state.getEncounterSlot(), gen.next<0, 10000>()));
+            break;
+        default:
+            break;
+        }
+
+        PersonalInfo info = PersonalLoader::getPersonal(Game::BDSP, encounterArea.getPokemon()[state.getEncounterSlot()].getSpecie(), 0);
```
Admiral-Fish, line 77: Each Slot in EncounterArea8 has PersonalInfo attached. That should be used instead

Source/Core/Gen8/Generators/WildGenerator8.cpp
```diff
@@ -18,26 +18,181 @@
  */
 
 #include "WildGenerator8.hpp"
+#include <Core/Enum/Game.hpp>
 #include <Core/Enum/Method.hpp>
+#include <Core/Parents/PersonalLoader.hpp>
```
Admiral-Fish, line 23: Remove after other change

Source/Core/Gen8/Generators/WildGenerator8.cpp
```diff
@@ -18,26 +18,181 @@
...
+        state.setAbility(gen.next() % 2);
+
+        if (encounterArea.getLocation() > 222 && encounterArea.getLocation() < 244)
+        { // TODO: add unown check
+            u32 unownForm = gen.next() % 28; // Form call
+        }
+
+        if (lead == Lead::CuteCharm)
+        {
+            if (info.getGender() + 2 > 2)
```
Admiral-Fish, line 122: What exactly is this checking for?

```
if ((info.getGender() + 2 > 2) && (gen.next() % 3)
```

Will help with some of the ugliness of all these if else

Source/Core/Gen8/Generators/WildGenerator8.cpp
```diff
@@ -18,26 +18,181 @@
...
+                else if (info.getGender() == 254)
+                {
+                    state.setGender(1);
+                }
+                else if (info.getGender() == 0)
+                {
+                    state.setGender(0);
+                }
+            }
+        }
```
Admiral-Fish, line 157: All of this can be removed after the above change

Source/Core/Gen8/Generators/WildGenerator8.cpp
SteveCookTU: Should be cleaned up now

Source/Core/Gen8/Generators/WildGenerator8.cpp
SteveCookTU: Still need the duplicate of the gender random for when lead isn't cute charm

SteveCookTU (approved):

PR #219, opened 2022-01-05
Move ability rand outside of if statement

SteveCookTU:


Source/Core/Gen8/Generators/StaticGenerator8.cpp
```diff
@@ -111,14 +111,12 @@ std::vector<StaticState> StaticGenerator8::generate(u64 seed0, u64 seed1, const
             }
         }
 
+        state.setAbility(rngList.getValue() % 2);
```
Admiral-Fish, line 114: No need for 2 calls to setAbility. Do something like

u8 ability = rand()
if (need to change because of template)
    ability = template.ability
state.setAbility(ability)

Admiral-Fish (changes requested):

Admiral-Fish (approved):

PR #222, opened 2022-01-20
[Gen 8] Fixed egg PID generation

ElliotGr:
The PID wasn't getting rolled without Masuda method or shiny charm leading to all PIDs being 0

### DIFF
--- a/Source/Core/Gen8/Generators/EggGenerator8.cpp
+++ b/Source/Core/Gen8/Generators/EggGenerator8.cpp
@@ -39,7 +39,7 @@ std::vector<EggState> EggGenerator8::generate(u64 seed0, u64 seed1) const
 
     RNGList<u32, Xorshift, 2, 0> rngList(rng);
 
-    u8 pidRolls = 0;
+    u8 pidRolls = 1;
     if (daycare.getMasuda())
     {
         pidRolls += 6;


SteveCookTU:
The problem is when masuda and/or shiny charm aren't being used, PID is created based on the main state (iirc) when you pick up the egg and not based on the 32bit seed in the game's memory. Also just adding one to the PID rerolls will throw off masuda and shiny charm PIDs as it will reroll an extra time

ElliotGr:
> The problem is when masuda and/or shiny charm aren't being used, PID is created based on the main state (iirc) when you pick up the egg and not based on the 32bit seed in the game's memory. Also just adding one to the PID rerolls will throw off masuda and shiny charm PIDs as it will reroll an extra time

Lol, just typed out a reply based on https://bulbapedia.bulbagarden.net/wiki/Masuda_method and then read the whole thing and realised you're right about the extra reroll because of a bug.  I'd say just change the +6 and +2 to +5 and +1 but sounds like it's not that simple! Sorry, bit of a noob at all this, was just throwing together a quick hack to make it work for me lol.

Admiral-Fish:
Thanks for the initiative here. I haven't got around to it yet but my plan is too simply just not show the PID/Shiny columns if shiny charm or masuda isn't selected.

PR #230, opened 2022-02-08
PLA Wild

SteveCookTU:


Source/Core/Gen8/Generators/WildGeneratorPLA.cpp
```diff
@@ -0,0 +1,142 @@
...
+        Xoroshiro mainRng(seed);
+        mainRng.next();
+        mainRng.next();
+        seed = mainRng.next();
+    }
+
+    for (u32 advance = 0; advance <= maxAdvances; advance++)
+    {
+        State result(initialAdvances + advance);
+        Xoroshiro mainRng(seed);
```
Admiral-Fish, line 31: Would making use of RNGList be possible here. We are using several different Xoroshiros here

Source/Forms/Gen8/WildPLA.cpp
```diff
@@ -0,0 +1,156 @@
...
+    setting.setValue("geometry", this->saveGeometry());
+    setting.endGroup();
+
+    delete ui;
+}
+
+bool WildPLA::hasProfiles() const
+{
+    return !profiles.empty();
+}
```
Admiral-Fish, line 37: Is this function necessary?

Source/Forms/Gen8/WildPLA.cpp
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    QSettings setting;
+    int val = setting.value("WildPLA/profile", 0).toInt();
+    if (val < ui->comboBoxProfiles->count())
+    {
+        ui->comboBoxProfiles->setCurrentIndex(val);
+    }
+
+    profilesIndexChanged(0);
```
Admiral-Fish, line 61: I don't think we want this here

Source/Forms/Gen8/WildPLA.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <QMenu>
+#include <QSettings>
+
+WildPLA::WildPLA(QWidget *parent) : QWidget(parent), ui(new Ui::WildPLA)
+{
+    ui->setupUi(this);
+
+    setAttribute(Qt::WA_QuitOnClose, false);
+
+    updateProfiles();
```
Admiral-Fish, line 19: Please move this into setupModels(), I recently changed all the others forms as well

Source/Forms/Gen8/WildPLA.cpp
```diff
@@ -0,0 +1,156 @@
...
+    {
+        ui->comboBoxProfiles->setCurrentIndex(val);
+    }
+
+    profilesIndexChanged(0);
+}
+
+void WildPLA::setupModels()
+{
+    model = new RaidModel(ui->tableView);
```
Admiral-Fish, line 66: Is this the model we want to be using?

Admiral-Fish (changes requested):

Source/Core/Gen8/Generators/WildGeneratorPLA.cpp
```diff
@@ -0,0 +1,142 @@
...
+                                   u8 rolls, bool isAlpha) :
+    Generator(initialAdvances, maxAdvances, tid, sid, genderRatio, Method::None, filter), rolls(rolls), isAlpha(isAlpha)
+{
+}
+
+std::vector<State> WildGeneratorPLA::generate(u64 seed) const
+{
+    std::vector<State> states;
+
+    seed -= 0x82A2B175229D6A5B;
```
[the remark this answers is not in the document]
SteveCookTU: Game does not store the previous "generator" seed so we do this to reverse the first spawner seed to get the previous and be able to also display the first however many advances.

Source/Core/Gen8/Generators/WildGeneratorPLA.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    for (u32 advance = 0; advance < initialAdvances; advance++)
+    {
+        // Probably horrible implementation but I forget how RNGList works
+        Xoroshiro mainRng(seed);
+        mainRng.next();
+        mainRng.next();
+        seed = mainRng.next();
+    }
```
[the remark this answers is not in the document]
SteveCookTU: This is how the game advances the generator seed which is used to advance the RNG

Source/Core/Enum/Game.hpp
```diff
@@ -59,7 +59,8 @@ enum class Game : u32
     BD = 1 << 26,
     SP = 1 << 27,
     BDSP = BD | SP,
-    Gen8 = SwSh | BDSP
+    Gen8 = SwSh | BDSP,
+    PAL = 1 << 28
```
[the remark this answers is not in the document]
SteveCookTU: It's still in the air whether the game should be considered Gen 8 and so far has just been treated as if it were. Can easily throw it in to include it

Source/Core/Gen8/Generators/WildGeneratorPLA.cpp
SteveCookTU: Honestly I didn't even think of using it because of all the Xoroshiro instances. Don't believe it would be easily possible as it needs to reseed the whole xoroshiro of the first to then create the rest again

Source/Forms/Gen8/WildPLA.cpp
SteveCookTU: No, will remove. Extra from copying over other form as a template

Source/Forms/Gen8/WildPLA.cpp
SteveCookTU: Was here from copying over from another form as a template, will remove.

Source/Forms/Gen8/WildPLA.cpp
SteveCookTU: We really can use wild8 or raid. I'm thinking with the potential advanced features that can be added it will eventually acquire its own model. The generation and logic of how the pokemon are generated are a copy and paste so thought it would be a temporary solute to raids

Source/Core/Gen8/Generators/WildGeneratorPLA.cpp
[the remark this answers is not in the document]
Admiral-Fish: Please leave a comment documenting this

Source/Core/Enum/Game.hpp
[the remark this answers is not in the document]
Admiral-Fish: Fair, don't forget to correct the typo

PR #233, opened 2022-02-19
[Gen 4] Bug Catching Contest (partial #1)

Real96:
Added BCC in Wild Generator and Searcher (partial #1)


Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -352,82 +353,62 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
...
+            break;
         case Encounter::HeadButt: // TODO
-        case Encounter::BugCatchingContest: // TODO
         default:
             break;
         }
 
         u32 pid = 0;
-        switch (lead)
+        for (u8 loop = 0; loop < loops; loop++)
```
Admiral-Fish, line 373: I'm not sure if this is the best approach. In the successful cute charm case, we skip the loop as well for bug contest

Might need to refactor this section to leave the non bug contest case the same and better handle bug contest

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -352,82 +353,62 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
...
             break;
+        case Encounter::BugCatchingContest:
+            loops = 4;
+            state.setEncounterSlot(EncounterSlot::kSlot(first, encounter));
+            if (!filter.compareEncounterSlot(state))
+            {
+                continue;
+            }
+
+            state.setLevel(0);
```
Admiral-Fish, line 364: Missing level calculation?

Source/Core/Gen4/Encounters4.cpp
```diff
@@ -249,6 +250,7 @@ namespace Encounters4
                 modifySwarmHGSS(grass, data, info, encounter, profile.getSwarm());
 
                 encounters.emplace_back(location, Encounter::Grass, grass);
+                encounters.emplace_back(location, Encounter::BugCatchingContest, bcc);
```
Admiral-Fish, line 253: Are we missing the data for this?

Source/Core/Gen4/Searchers/WildSearcher4.cpp
```diff
@@ -667,6 +667,11 @@ bool WildSearcher4::encounterMethodK(WildState &state, u32 seed) const
         state.setLevel(encounterArea.calcLevel(state.getEncounterSlot(), seed >> 16));
         state.setSeed(rng.next());
         break;
+    case Encounter::BugCatchingContest:
+        state.setEncounterSlot(EncounterSlot::kSlot(rng.nextUShort(), encounter));
+        state.setLevel(0);
```
Admiral-Fish, line 672: Missing level calculation?

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -296,12 +297,32 @@ void Wild4::search()
...
+        {
+            if (min[i] == 31)
+            {
+                flag = false;
+                break;
+            }
+        }
+        if (flag)
+        {
+            QMessageBox err(QMessageBox::Warning, ("Missing Flawless IV"), ("BCC Searcher needs at least one IV at 31"));
```
Admiral-Fish, line 317: Wrap the strings in tr() calls. Also use "Bug Contest" instead of BCC

Admiral-Fish (changes requested):

Source/Core/Gen4/Generators/WildGenerator4.cpp
SteveCookTU: Rather than duplicating a ton of lines for contest vs non-contest is place a loop within each section that would actually loop. More lines than one loop but more lines that duplicating the whole switch statement.

Real96: > I'm not sure if this is the best approach. In the successful cute charm case, we skip the loop as well for bug contest
> 
> Might need to refactor this section to leave the non bug contest case the same and better handle bug contest

How do we skip the loop in the cutecharm case? The if should avoid this

Real96: > Rather than duplicating a ton of lines for contest vs non-contest is place a loop within each section that would actually loop. More lines than one loop but more lines that duplicating the whole switch statement.

I don't think you can have multiple loop section, cause you have to loop until you find an IV at 31, max 4 times. I think you can't separate the loop

Source/Forms/Gen4/Wild4.cpp
Real96: done

Source/Core/Gen4/Generators/WildGenerator4.cpp
Real96: > I'm not sure if this is the best approach. In the successful cute charm case, we skip the loop as well for bug contest
> 
> Might need to refactor this section to leave the non bug contest case the same and better handle bug contest

Done

Source/Core/Gen4/Encounters4.cpp
```diff
@@ -81,6 +81,24 @@ namespace Encounters4
             return encounters;
         }
 
+        std::vector<std::vector<u8>> getBugCatchingContestData()
+        {
+
```
Admiral-Fish, line 86: nit: whitespace

Source/Core/Gen4/Encounters4.cpp
```diff
@@ -226,6 +244,25 @@ namespace Encounters4
             }
         }
 
+        std::vector<EncounterArea4> getBugCatchingContest(const std::vector<u8> &data, const PersonalInfo *info)
```
Admiral-Fish, line 247: This needs to be passed profile so we can include/exclude based on national dex

Source/Core/Gen4/Encounters4.cpp
```diff
@@ -410,11 +447,24 @@ namespace Encounters4
         auto *info = PersonalLoader::getPersonal(version);
 
         std::vector<EncounterArea4> encounters;
-        for (const auto &data : getData(version))
+
+        if (encounter == Encounter::BugCatchingContest)
```
Admiral-Fish, line 451: Only for hgss, include a game check

Source/Core/Gen4/Generators/WildGenerator4.cpp
Admiral-Fish: What was the conclusion here?

Admiral-Fish (changes requested):
We need to add national dex as a setting to Profile4

Source/Core/Gen4/Generators/WildGenerator4.cpp
SteveCookTU: Refactored to remove duplication. Don't think extra functions are necessary. It may look more readable but the effect is still the same so I wouldn't say it's high priority.

Source/Core/Gen4/Encounters4.cpp
Real96: Done

Real96:
> We need to add national dex as a setting to Profile4

Done

Source/Core/Parents/ProfileLoader.cpp
```diff
@@ -78,11 +78,11 @@ namespace ProfileLoader3
     {
         Profile3 getProfile(const json &j)
         {
-            std::string name = j["name"].get<std::string>();
-            Game version = j["version"].get<Game>();
-            u16 tid = j["tid"].get<u16>();
-            u16 sid = j["sid"].get<u16>();
-            bool deadBattery = j["battery"].get<bool>();
+            std::string name = j.value("name", std::string("-"));
```
Admiral-Fish, line 81: Two question for this entire file.

1. I think it is unnecessary to change to .value() for fields that already exist
2. What is up with all the c-style casts? Can the templated function not pick up on the type of the default value?

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -296,12 +297,32 @@ void Wild4::search()
     searcherModel->clearModel();
     searcherModel->setMethod(method);
```
Admiral-Fish, line 298: These need to be moved after the 31IV check

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -254,6 +254,7 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
     u8 buffer = 0;
     u8 thresh = 0;
     u16 rate = encounterArea.getEncounterRate();
+    u8 loops = 1;
```
Admiral-Fish, line 257: I am still not sure I like how this is handled. It seems very invasive to handle one special case. I think the core logic can be extracted to a function and the different encounters use this function as necessary

Source/Core/Parents/ProfileLoader.cpp
SteveCookTU: The casts are the same reason for the .get() from before. It uses the type of the second parameter when parsing. I understand some may not need the casting but I wasn't sure if C++ handled type safety like that automatically.

SteveCookTU: Changing it for all of them just provides consistency. I understand some will never be missing unless deleted by the user but the safety is better to have than to not.

Admiral-Fish: I removed all the type casting and it built fine. I suspect the fact that the type is on the left hand of the assignment is enough to inform the compiler of the templated types.

Real96: > I removed all the type casting and it built fine. I suspect the fact that the type is on the left hand of the assignment is enough to inform the compiler of the templated types.

removed all the type casting

Source/Forms/Gen4/Wild4.cpp
Real96: done

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -254,6 +408,7 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
...
```
Admiral-Fish, line 411: Leftover variable. Delete

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
     return states;
 }
 
+template <bool bug>
+void createPokemon(PokeRNG &go, WildState4 &state, u8 buffer, u8 synchNature, u8 genderRatio, u16 tsv, const Lead &lead,
```
Admiral-Fish, line 248: This should return a bool. False if the filter dictates it otherwise true, so the generateMethodK function can utilize this information

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+                // Get hunt nature
+                state.setNature(go.nextUShort<true>() % 25);
+
+                if (!filter.compareNature(state))
+                {
+                    return;
+                }
+
+                cuteCharmSuccess = true;
+                pid = buffer + state.getNature();
```
Admiral-Fish, line 269: Add a "return true" statement here. Cute charm bypasses the for loop below.

Just make sure to add IVs here as well.

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
 }
 
+template <bool bug>
+void createPokemon(PokeRNG &go, WildState4 &state, u8 buffer, u8 synchNature, u8 genderRatio, u16 tsv, const Lead &lead,
+                   const StateFilter &filter)
+{
+    u32 pid = 0;
+    if constexpr (bug)
+    {
+        bool cuteCharmSuccess = false;
```
Admiral-Fish, line 254: This variable shouldn't be necessary with the below change to the true section of cute charm

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+                state.setLead(Lead::CompoundEyes);
+                break;
+            default:
+                break;
+            }
+
+            if (!cuteCharmSuccess)
+            {
+                // Get hunt nature
+                state.setNature(go.nextUShort<true>() % 25);
```
Admiral-Fish, line 297: This can get moved to the default case of the switch statement above

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+                }
+                break;
+            case Lead::CompoundEyes:
+                state.setLead(Lead::CompoundEyes);
+                break;
+            default:
+                break;
+            }
+
+            if (!cuteCharmSuccess)
```
Admiral-Fish, line 294: Can get rid of this if statement with the other changes to cute charm

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+            state.setGender(pid & 255, genderRatio);
+            state.setShiny<8>(tsv, (pid >> 16) ^ (pid & 0xffff));
+
+            u16 iv1 = go.nextUShort<true>();
+            u16 iv2 = go.nextUShort<true>();
+
+            state.setIVs(iv1, iv2);
+            state.calculateHiddenPower();
+
+            if constexpr (bug)
```
Admiral-Fish, line 319: This if statement isn't necessary. We are already in a "if constexpr (bug)" section

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+            }
+            break;
+        default:
+            state.setNature(go.nextUShort<true>() % 25);
+            break;
+        }
+
+        if (!filter.compareNature(state))
+        {
+            return;
```
Admiral-Fish, line 377: return false

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+        state.setGender(pid & 255, genderRatio);
+        state.setShiny<8>(tsv, (pid >> 16) ^ (pid & 0xffff));
+
+        u16 iv1 = go.nextUShort<true>();
+        u16 iv2 = go.nextUShort<true>();
+
+        state.setIVs(iv1, iv2);
+        state.calculateHiddenPower();
+    }
+}
```
Admiral-Fish, line 399: Add a return true here at the end of the function

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+            || lead == Lead::CuteCharm875M)
+        {
+            if ((go.nextUShort<true>() % 3) != 0) // Successfull cute charm
+            {
+                // Get hunt nature
+                state.setNature(go.nextUShort<true>() % 25);
+
+                if (!filter.compareNature(state))
+                {
+                    return;
```
Admiral-Fish, line 265: return false

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,160 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+                }
+
+                cuteCharmSuccess = true;
+                pid = buffer + state.getNature();
+            }
+        }
+
+        for (u8 loop = 0; loop < 4; loop++)
```
Admiral-Fish, line 273: After this for loop add a nature filter check

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -308,6 +463,7 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
 
             state.setLevel(encounterArea.calcLevel(state.getEncounterSlot()));
             occidentary += 1; // Compensate for the game's advances after the battle ends
+            createPokemon<false>(go, state, buffer, synchNature, genderRatio, tsv, lead, filter);
```
Admiral-Fish, line 466: Use returned bool to continue or not

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -318,6 +474,7 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
 
             state.setLevel(encounterArea.calcLevel(state.getEncounterSlot(), go.nextUShort<true>()));
             occidentary += 1; // Compensate for the game's advances after the battle ends
+            createPokemon<false>(go, state, buffer, synchNature, genderRatio, tsv, lead, filter);
```
Admiral-Fish, line 477: Use returned bool to continue or not

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -336,6 +493,7 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
             state.setLevel(encounterArea.calcLevel(state.getEncounterSlot()));
             occidentary += 2; // Compensate for the game's advances after the battle ends
             go.next();
+            createPokemon<false>(go, state, buffer, synchNature, genderRatio, tsv, lead, filter);
```
Admiral-Fish, line 496: Use returned bool to continue or not

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -351,109 +509,27 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
 
             state.setLevel(encounterArea.calcLevel(state.getEncounterSlot(), go.nextUShort<true>()));
             occidentary += 1; // Compensate for the game's advances after the battle ends
+            createPokemon<false>(go, state, buffer, synchNature, genderRatio, tsv, lead, filter);
```
Admiral-Fish, line 512: Use returned bool to continue or not

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -351,109 +509,27 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
...
-            // Begin search for valid pid
-            do
-            {
-                u16 low = go.nextUShort<true>();
-                u16 high = go.nextUShort<true>();
-                pid = static_cast<u32>((high << 16) | low);
-            } while (pid % 25 != state.getNature());
-
+            state.setLevel(encounterArea.calcLevel(state.getEncounterSlot(), go.nextUShort<true>()));
+            createPokemon<true>(go, state, buffer, synchNature, genderRatio, tsv, lead, filter);
```
Admiral-Fish, line 522: Use returned bool to continue or not

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -351,109 +509,27 @@ std::vector<WildState4> WildGenerator4::generateMethodK(u32 seed, const Encounte
...
-
-        u16 iv1 = go.nextUShort<true>();
-        u16 iv2 = go.nextUShort<true>();
-
-        state.setIVs(iv1, iv2);
-        state.calculateHiddenPower();
+        if (encounter == Encounter::BugCatchingContest && !filter.compareNature(state))
+        {
+            continue;
+        }
```
Admiral-Fish, line 532: This shouldn't be necessary with the other changes

Admiral-Fish (changes requested):

Source/Core/Gen4/Generators/WildGenerator4.cpp
SteveCookTU: It still has to run through one iteration of the loop to generate the IVs, nature, etc. We can have it either call createPokemon recursively but with the bool as false (but then it would check chute charm again) or just reintroduce the loops variable and change it to 1 instead of 4 in the case of successful cute charm.

Source/Core/Gen4/Generators/WildGenerator4.cpp
```diff
@@ -244,6 +244,147 @@ std::vector<WildState4> WildGenerator4::generateMethodJ(u32 seed, const Encounte
...
+            u16 iv2 = go.nextUShort<true>();
+
+            state.setIVs(iv1, iv2);
+
+            for (u8 iv = 0; iv < 6; iv++)
+            {
+                if (state.getIV(iv) == 31)
+                {
+                    loop = 4;
+                    break;
```
Admiral-Fish, line 322: Would it make more sense to just `return true;` here?

Admiral-Fish (approved):

PR #242, opened 2022-03-02
[Gen 4] Hide Suction Cups lead ability with gen4 non-fishing encounters

Real96:


Source/Forms/Gen4/Wild4.cpp
```diff
@@ -435,6 +426,16 @@ void Wild4::generatorEncounterIndexChanged(int index)
             break;
         }
 
+        if ((currentProfile->getVersion() & Game::HGSS) != Game::None && index > 2)
```
Admiral-Fish, line 429: Change index to using the encounter type we get above. Will be easy to read when mark what encounters we care about instead of wondering what does "index > 2" mean

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -435,6 +426,16 @@ void Wild4::generatorEncounterIndexChanged(int index)
...
+        {
+            ui->toolButtonGeneratorLead->removeAction(tr("Suction Cups"));
+            ui->toolButtonGeneratorLead->addAction(tr("Suction Cups"), toInt(Lead::SuctionCups));
```
Admiral-Fish, line 432: Not the biggest fan of this approach. Should check if it exists and if not then add it. This check should be handled internally to the ComboMenu class

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -465,6 +466,16 @@ void Wild4::searcherEncounterIndexChanged(int index)
             break;
         }
 
+        if ((currentProfile->getVersion() & Game::HGSS) != Game::None && index > 2)
+        {
+            ui->comboBoxSearcherLead->removeItem(ui->comboBoxSearcherLead->findData(toInt(Lead::SuctionCups)));
+            ui->comboBoxSearcherLead->addItem(tr("Suction Cups"), toInt(Lead::SuctionCups));
```
Admiral-Fish, line 472: Same deal as above. Only add if it doesn't exist

Admiral-Fish (changes requested):

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -465,6 +466,16 @@ void Wild4::searcherEncounterIndexChanged(int index)
             break;
         }
 
+        if ((currentProfile->getVersion() & Game::HGSS) != Game::None && index > 2)
```
Admiral-Fish, line 469: Change index to using the encounter type we get above. Will be easy to read when mark what encounters we care about instead of wondering what does "index > 2" mean

Source/Forms/Gen4/Wild4.cpp
Real96: done

Source/Forms/Gen4/Wild4.cpp
Real96: done

Source/Forms/Gen4/Wild4.cpp
Real96: done

Source/Forms/Gen4/Wild4.cpp
Real96: done

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -435,6 +426,18 @@ void Wild4::generatorEncounterIndexChanged(int index)
...
         }
 
+        if ((currentProfile->getVersion() & Game::HGSS) != Game::None && (encounter == Encounter::OldRod || encounter == Encounter::GoodRod || encounter == Encounter::SuperRod))
+        {
+            if (!ui->toolButtonGeneratorLead->findAction(tr("Suction Cups")))
+            {
+                ui->toolButtonGeneratorLead->addAction(tr("Suction Cups"), toInt(Lead::SuctionCups));
+            }
+        }
+        else if (ui->toolButtonGeneratorLead->findAction(tr("Suction Cups")))
```
Admiral-Fish, line 436: I don't think the check here is necessary. Remove already internally checks if the item exists

Source/Forms/Gen4/Wild4.cpp
```diff
@@ -465,6 +468,18 @@ void Wild4::searcherEncounterIndexChanged(int index)
...
         }
 
+        if ((currentProfile->getVersion() & Game::HGSS) != Game::None && (encounter == Encounter::OldRod || encounter == Encounter::GoodRod || encounter == Encounter::SuperRod))
+        {
+            if (ui->comboBoxSearcherLead->findData(toInt(Lead::SuctionCups)) < 0)
+            {
+                ui->comboBoxSearcherLead->addItem(tr("Suction Cups"), toInt(Lead::SuctionCups));
+            }
+        }
+        else if (ui->comboBoxSearcherLead->findData(toInt(Lead::SuctionCups)) >= 0)
```
Admiral-Fish, line 478: I don't think the check here is necessary. Remove doesn't do anything if the index is out of range, so we would be calling findData twice when we only need to once.

Admiral-Fish (changes requested):

Source/Forms/Gen4/Wild4.cpp
Real96: reverted

Source/Forms/Gen4/Wild4.cpp
Real96: reverted

PR #251, opened 2022-04-11
[Gen 8] Add gender display for Nidorant/Volbeat/Indeedee

SteveCookTU:


Source/Core/Gen8/Generators/EggGenerator8.cpp
```diff
@@ -72,25 +72,31 @@ std::vector<EggState> EggGenerator8::generate(u64 seed0, u64 seed1) const
...
-                state.setGender(1);
-            }
-            else if (genderRatio == 0)
-            {
-                state.setGender(0);
+                state.setGender(gen.next(2));
+                if (genderRatio != 255 && genderRatio != 254 && genderRatio != 0)
+                {
+                    gen.next();
+                }
```
Admiral-Fish, line 79: Why is this part necessary?

Source/Core/Gen8/Generators/EggGenerator8.cpp
```diff
@@ -72,25 +72,31 @@ std::vector<EggState> EggGenerator8::generate(u64 seed0, u64 seed1) const
             {
                 // gen.next(2);
                 // Handle display result later
```
Admiral-Fish, line 74: If these comments are no longer relevant please remove them

Admiral-Fish (changes requested):

Source/Core/Gen8/Generators/EggGenerator8.cpp
SteveCookTU: Based on previous logic, the next rand call was still affected based on the gender ratio. What would the expected gender ratio be in the filters for nidoran volbeat? A 100% one/genderless or 50/50?

Admiral-Fish: Is it accurate to what the game does?

SteveCookTU: Will need to double check. It's more we should be seeing if the game treats them as 50/50 and has an extra gender call or treats them as their base 100% and only has the one.

Admiral-Fish:
This is accounted for in the redesign branch

PR #259, opened 2022-07-23
Allow omitting nature in IV calc

Lincoln-LM:
This PR serves to allow omitting nature in IV calculation for easier calcing of shadow pokemon's IVs pre-purification

This is a pretty simple change, so if there's a different way you want to go about doing this/any other comments lmk


Admiral-Fish:
Will this be useful after the redesign branch? I am adding showing stats

Lincoln-LM:
I'd argue it's less useful with showing + filtering by stats implemented, but unless iv calc is entirely removed its a simple change thats still an improvement allows you to also use it with shadows (or in a scenario where you can only see stats in battle and not nature).

Admiral-Fish:
I rebased and got it working with the changes I have made to IV Calculator over time. The only thing I am not sure about is the "Next Level" stuff, I'm not sure that can work with an unknown nature

PR #277, opened 2022-10-26
Add std:: namespace to size_t

bitofhope:
Compilation fails against tag v4.0.0 on gcc 10.2.1. This change fixed build on origin/master with my other laptop about an hour ago, but it looks like 1edea7f935e14f29066f2d309752b6eda4b36d2e broke the build for me on this other machine.

Didn't look too deep into the latter issue yet, but this seems to be one of the things the build complains about, anyway.


Admiral-Fish:
I had included this as part of #289

PR #284, opened 2022-12-07
Fix linux compilation

GuillaumeCisco:
same as https://github.com/Admiral-Fish/PokeFinder/pull/277


Admiral-Fish:
I had included this as part of #289

PR #292, opened 2023-04-25
Fix build with GCC 12 on Arch Linux

### LINKED ISSUE #291: Build fails because `size_t` was not found in scope
While compiling `v4.1.0` for the Arch User Repository, the build fails with the following error message:

```
In file included from /.../Source/Core/Gen3/Encounters3.cpp:20:
/.../Source/Core/Gen3/Encounters3.hpp:59:42: error: ‘size_t’ was not declared in this scope; did you mean ‘std::size_t’?
   59 |     const ShadowTemplate *getShadowTeams(size_t *size = nullptr);
      |                                          ^~~~~~
      |                                          std::size_t
```

I assume the type should be `std::size_t`, not `size_t`? There's at least one more occurrence of the same type further down the file and in the corresponding `.c` file.

Summary of CMake options:
```
-- Summary of build options:

    Package version: 1.1.0
    Library version: 1:9:0
    Install prefix:  /usr/local
    Target system:   Linux
    Compiler:
        Build type:     RELEASE
        C compiler:     /usr/lib/ccache/bin/cc
        CFLAGS:         -O3   -march=native -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection -flto=auto
        WARNCFLAGS:      -Wall -Wextra -Wmissing-prototypes -Wstrict-prototypes -Wmissing-declarations -Wpointer-arith -Wdeclaration-after-statement -Wformat-security -Wwrite-strings -Wshadow -Winline -Wnested-externs -Wfloat-equal -Wundef -Wendif-labels -Wempty-body -Wcast-align -Wclobbered -Wvla -Wpragmas -Wunreachable-code -Waddress -Wattributes -Wdiv-by-zero -Wconversion -Wformat-nonliteral -Wmissing-field-initializers -Wmissing-noreturn -Wsign-conversion -Wunused-macros -Wunused-parameter -Wredundant-decls -Wno-format-nonliteral
    Test:
        Python:         TRUE (3.10.10, /usr/bin/python3.10)
    Docs:
        Build docs:     OFF
    Features:
        Applications:   OFF
        Examples:       OFF
```

Compiler is GCC 12.2.1:

```shellsession
$ /usr/lib/ccache/bin/cc --version
cc (GCC) 12.2.1 20230201
```

phijor:
For reasons beyond my understanding, PokeFinder failed to build on Arch Linux with GCC 12. The errors indicated wrong/missing imports, which I fixed:

1. `size_t` not in scope: Qualify all occurrences as `std::size_t` (fixes #291)
2. `std::find` not found  (lol): Include the `algorithm` header, where it is declared

These should be rather conservative changes that hopefully don't break anything else.

### DIFF
--- a/Source/Core/Gen3/Encounters3.cpp
+++ b/Source/Core/Gen3/Encounters3.cpp
@@ -74,7 +74,7 @@ namespace Encounters3
     std::vector<EncounterArea3> getEncounters(Encounter encounter, Game version)
     {
         const u8 *compressedData;
-        size_t compressedLength;
+        std::size_t compressedLength;
 
         if (version == Game::Emerald)
         {
@@ -108,7 +108,7 @@ namespace Encounters3
         const PersonalInfo *info = PersonalLoader::getPersonal(version);
 
         std::vector<EncounterArea3> encounters;
-        for (size_t offset = 0; offset < length; offset += sizeof(WildEncounter3))
+        for (std::size_t offset = 0; offset < length; offset += sizeof(WildEncounter3))
         {
             const auto *entry = reinterpret_cast<const WildEncounter3 *>(data + offset);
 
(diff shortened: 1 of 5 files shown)


Admiral-Fish:
I opted to make it match how Gen 8 is using `int` instead of `size_t`. Thanks for bringing this issue to my attention

PR #296, opened 2023-05-13
Fix TSV and fake sidtid masking for BDSP

SteveCookTU:


Admiral-Fish:
Does this need test updates?

SteveCookTU:
> Does this need test updates?

All current tests still pass with the changes

PR #299, opened 2023-05-26
Fix build

sagehane:
Here's some changes I needed to apply to get PokeFinder compiling on NixOS (and presumably any other system, idk how it's not causing issues for the CI).

### Notes

The README has only been updated for the Linux section because that's what I tested it for. It should be the same on Windows and macOS but I didn't care to check. Maybe updating the README should be left to a different PR.

The build system currently doesn't seem to check if the `requests` Python package is installed, and it was a bit hard to tell when/why the CMake step failed.

The current build system seems to require network access at build-time, which seems to cause issues for packaging PokeFinder on distros like NixOS. I don't really know enough Nix magic to work around it through Nix's side, so it's likely that the package on Nixpkg will stay stuck at version 4.0.1 for a while.

### Reproducing the build issue
One should be able to reproduce the issue I hit by following these steps on the current master branch of the repo:
1. Install a version of Nix capable of using Nix Flakes
2. Copy the `flake.nix` file below to the root dir of the project
3. Run `nix develop` to create an env with all the dependencies installed
4. Follow the instructions on https://github.com/Admiral-Fish/PokeFinder#building

flake.nix:
```nix
# Currently doesn't build with `nix build .?submodules=1#` because the build
# system seems to require network connection at build-time

{
  inputs.flake-utils.url = "github:numtide/flake-utils";

  outputs = { self, nixpkgs, flake-utils }:
    flake-utils.lib.eachDefaultSystem (system:
      let pkgs = nixpkgs.legacyPackages.${system}; in
      {
        packages.default = with pkgs; stdenv.mkDerivation {
            name = "PokeFinder";
            src = self;

            nativeBuildInputs = [
              cmake
              qt6.wrapQtAppsHook
            ] ++ (with python3Packages; [
              pytest # Optional
              requests
            ]);

            buildInputs = with qt6; [
              qttools
              qtwayland
            ];

            installPhase = ''
              install -D Source/Forms/PokeFinder $out/bin/PokeFinder
            '';
          };
      }
    );
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ Linux
 - Install Qt 6
   - [Qt website](https://www.qt.io/download)
   - sudo apt install qt6-base-dev
+- Install Python 3 and `requests`
+  - python -m pip install requests
 - Download the linux zip folder from the [releases page](https://github.com/Admiral-Fish/PokeFinder/releases/latest)
 - Extract PokéFinder from the zip folder
 
@@ -85,7 +87,7 @@ Windows
   - cmake --build .
 - Bundle
   - mk PokeFinder-windows
-  - move release\PokeFinder.exe PokeFinder-windows\PokeFinder.exe 
+  - move release\PokeFinder.exe PokeFinder-windows\PokeFinder.exe
   - windeployqt --release --no-translations --no-angle --no-plugins --no-opengl-sw PokeFinder.exe
   - xcopy /I "QTPath"\plugins\platforms\qwindows.dll PokeFinder-windows\platforms\
   - xcopy /I "QTPath"\plugins\styles\qwindowsvistastyle.dll PokeFinder-windows\styles\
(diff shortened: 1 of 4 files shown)


sagehane:
I just realised that this is a duplicate of #292, although the approach taken to address the issue is slightly different.
I think adding `#include <cstddef>` is more consistent as that pattern is seen in other files too.

Admiral-Fish:
Should be fixed now. I reworked the build to not rely on `requests` anymore.

sagehane:
I can confirm that the master branch now builds properly. Thanks.

Here's the updated `flake.nix` file I used to get it compiling:
```nix
# Build with `nix build .?submodules=1#`
# Get a shell with dependencies with `nix develop`

{
  inputs.flake-utils.url = "github:numtide/flake-utils";

  outputs = { self, nixpkgs, flake-utils }:
    flake-utils.lib.eachDefaultSystem (system:
      let pkgs = nixpkgs.legacyPackages.${system}; in
      {
        packages.default = with pkgs; stdenv.mkDerivation {
          name = "PokeFinder";
          src = self;

          nativeBuildInputs = [
            cmake
            qt6.wrapQtAppsHook

            # Optional
            #python3Packages.pytest # For testing
          ];

          buildInputs = with qt6; [
            qttools
            qtwayland
          ];

          installPhase = ''
            install -D Source/PokeFinder $out/bin/PokeFinder
          '';
        };
      }
    );
}
```

PR #307, opened 2023-06-12
Fix Gen5 IDs advances off by 1

SteveCookTU:


Real96:
[This](d8ae914fa2927abba0d2c5f0764d9e2bca92f100) was the fix before the redesign was merged

PR #344, opened 2023-10-21
BDSP egg star/square shininess fix

Lincoln-LM:
Needs testing but the logic is simple


Lincoln-LM:
CI fails on tests

Admiral-Fish:
Tests will need to be updated before I can merge this

PR #398, opened 2024-06-12
Allow IV filters to ignore inherited stats

c-poole:
Adds a third filter class DaycareFilter that each of the egg tabs now use, which has the option to ignore inherited stats for the purposes of filtering the results. 

The primary utility I see for this feature is to make it so that searches for targets do not explicitly depend on the provided values of the parent IVs and instead provides a blueprint for what the parent IVs need to be to match the filters. This allows for a workflow pattern of identifying a target -> identifying parent targets to allow you to hit said target -> getting target parents/egg.  

It also has some utility in gen 3 where there is an incomplete enumeration of the possible VBlank signatures, and therefore even if the user provides all the correct information for an egg they produced,  they may not get a hit on any of the provided methods.

### DIFF
--- a/Source/Core/Parents/Filters/StateFilter.cpp
+++ b/Source/Core/Parents/Filters/StateFilter.cpp
@@ -19,6 +19,8 @@
 
 #include "StateFilter.hpp"
 #include <Core/Parents/States/WildState.hpp>
+#include <Core/Parents/States/EggState.hpp>
+
 
 StateFilter::StateFilter(u8 gender, u8 ability, u8 shiny, bool skip, const std::array<u8, 6> &min, const std::array<u8, 6> &max,
                          const std::array<bool, 25> &natures, const std::array<bool, 16> &powers) :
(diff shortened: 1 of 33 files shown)


c-poole:
Is this feature not desired? I personally often want to use the egg page as a way to both find egg targets but also identify what parents need to be to reach desired goals, which this facilitates better options for. It is an additional "thing for the user to think about" which may not be desired. Can definitely adjust implementation as needed if it is desired.

Admiral-Fish:
I get side tracked easily. I'll try to look this weekend at this

Source/Core/Parents/Filters/StateFilter.cpp
```diff
@@ -219,3 +221,85 @@ bool WildStateFilter::compareState(const WildState &state) const
...
+
+    if (shiny != 255 && !(shiny & state.getShiny()))
+    {
+        return false;
+    }
+
+
+    for (int i = 0; i < 6; i++)
+    {
+
```
Admiral-Fish, line 266: remove space

Source/Form/Controls/Filter.hpp
```diff
@@ -94,22 +112,25 @@ class Filter : public QWidget
...
-        if constexpr (wild)
-        {
+
+        if constexpr(type == 1){
             return FilterType(getGender(), getAbility(), getShiny(), getDisableFilters(), getMinIVs(), getMaxIVs(), getNatures(),
                               getHiddenPowers(), getEncounterSlots());
         }
-        else
-        {
+        else if constexpr(type == 2){
```
Admiral-Fish, line 127: type == FilterType::Daycare

PR #399, opened 2024-06-13
Make Magnet Pull and Static apply to correct encounter types in Gen 3

c-poole:
Magnet Pull should only be applied for Grass encounters.
Static should only be applied for Grass and Water encounters.

This can be seen from the following PokeEmerald code:
https://github.com/pret/pokeemerald/blob/18f84b78f2d1a8669753fa586836fca06036c790/src/wild_encounter.c#L422-L446
https://github.com/pret/pokeemerald/blob/18f84b78f2d1a8669753fa586836fca06036c790/src/wild_encounter.c#L458-L465

Current behavior is that they both apply to all wild encounter types. Fix involves modifying the wild generator and searcher classes to recognize when they should be skipped. Alternative solution would be to reset selected Lead to None if either of these were selected and to remove the option to select them when the encounter type is changed to an incompatible type.

### DIFF
--- a/Source/Form/Gen3/Wild3.cpp
+++ b/Source/Form/Gen3/Wild3.cpp
@@ -197,37 +197,27 @@ void Wild3::generatorEncounterIndexChanged(int index)
         {
         case Encounter::Grass:
             ui->filterGenerator->setEncounterSlots(12);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), false);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), false);
             break;
         case Encounter::RockSmash:
-            ui->filterGenerator->setEncounterSlots(5);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), true);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), true);
-            break;
         case Encounter::Surfing:
+        case Encounter::SuperRod:
             ui->filterGenerator->setEncounterSlots(5);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), false);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), true);
             break;
         case Encounter::OldRod:
             ui->filterGenerator->setEncounterSlots(2);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), true);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), true);
             break;
         case Encounter::GoodRod:
             ui->filterGenerator->setEncounterSlots(3);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), true);
-            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), true);
-            break;
-        case Encounter::SuperRod:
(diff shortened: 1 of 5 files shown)


Admiral-Fish:
I think it makes more sense to handle this in the constructor

c-poole:
By this do you mean, scrubbing the lead passed into the constructor of the generator and searcher classes while still allowing the user to have leads that would be ignored selected or to actually modify the menu in response to the selected encounter type?

Admiral-Fish:
Yeah scrub out the lead in the constructor if it wouldn't be allowed for the encounter type

c-poole:
I've pushed a version that fixes this through a UI-behavior change that modifies the visibility of the Lead options according to what is applicable to the encounter type selected whenever it is changed. I'm fine with doing the constructor based approach if you still prefer that to this, but I think this is most transparent both from a code readability (someone reading through generator/searcher without realizing there is lead scrubbing would be very confused) and an end-user perspective. This change also fixes what I think are two bugs in the ComboMenu class HideAction method which automatically un-selects an Action corresponding to the provided data even if the method was told to set (keep) its visibility to true. The other is that it would hide a Menu if any of the Actions in the Menu are invisible instead of, as the comment suggests is desired, when all Actions are invisible.

Source/Form/Gen3/Wild3.cpp
```diff
@@ -197,22 +197,37 @@ void Wild3::generatorEncounterIndexChanged(int index)
...
+        case Encounter::SuperRod:
+            ui->filterGenerator->setEncounterSlots(5);
+            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), true);
+            ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), true);
             break;
         default:
             break;
         }
-
         updateEncounterGenerator();
```
Admiral-Fish, line 231: I think it will probably be cleaner to do it all below the switch statement. It also matches how I have done it in other places (for example Wild4.cpp)

```
bool magnet = encounter == Encounter::Grass;
bool static = encounter == Encounter::Grass || encounter == Encounter::Surfing;

ui->comboMenuGeneratorLead->hideAction(toInt(Lead::MagnetPull), !magnet);
ui->comboMenuGeneratorLead->hideAction(toInt(Lead::Static), !static);
```

Source/Form/Gen3/Wild3.cpp
```diff
@@ -399,24 +414,37 @@ void Wild3::searcherEncounterIndexChanged(int index)
...
             break;
         case Encounter::SuperRod:
             ui->filterSearcher->setEncounterSlots(5);
+            ui->comboMenuSearcherLead->hideAction(toInt(Lead::Static), true);
+            ui->comboMenuSearcherLead->hideAction(toInt(Lead::MagnetPull), true);
             break;
         default:
             break;
         }
-
```
Admiral-Fish, line 419: Same as other comment

Admiral-Fish (changes requested):
Nice catch on the combo menu stuff. Totally agree there

Source/Form/Gen3/Wild3.cpp
c-poole: Changed as desired but with different variable names since static is a keyword

Source/Form/Gen3/Wild3.cpp
c-poole: Changed to accommodate same as above

PR #405, opened 2024-07-18
Fix doubling up actions in PRs

SteveCookTU:
This PR works to help solve actions running twice for PR pushes. Solution found and described here: https://github.com/orgs/community/discussions/57827


SteveCookTU:
Another option is we could filter out the push actions (which may be better for the action lists) so that is only builds on pushes to master and PRs would take the pull_request event like intended

Issue #413, opened 2024-08-09
Manaphy seed search will never give results if shiny filter is active


NickPlayeZ:
When trying to search for a shiny Manaphy with a given TID/SID it’ll never show any results (probably due to the shiny lock).
However 99% of people that wanna rng this Manaphy to be shiny likely know how this works and to avoid confusion there should be the option (or it should be the default) to allow for the searcher to find shinies with the given TID/SID.

PR #414, opened 2024-08-09
Add shiny lock option to Event Manaphy research

Real96:
Add what asked in #413

### DIFF
--- a/Source/Form/Gen4/Static4.cpp
+++ b/Source/Form/Gen4/Static4.cpp
@@ -75,9 +75,6 @@ Static4::Static4(QWidget *parent) : QWidget(parent), ui(new Ui::Static4)
     connect(seedToTime, &QAction::triggered, this, &Static4::seedToTime);
     ui->tableViewSearcher->addAction(seedToTime);
 
-    ui->checkBoxGeneratorShinyLock->setVisible(true);
-    ui->checkBoxSearcherShinyLock->setVisible(true);
-
     connect(ui->pushButtonGenerate, &QPushButton::clicked, this, &Static4::generate);
     connect(ui->pushButtonSearch, &QPushButton::clicked, this, &Static4::search);
     connect(ui->pushButtonProfileManager, &QPushButton::clicked, this, &Static4::profileManager);
--- a/Source/Form/Gen4/Static4.cpp
+++ b/Source/Form/Gen4/Static4.cpp
@@ -75,6 +75,9 @@ Static4::Static4(QWidget *parent) : QWidget(parent), ui(new Ui::Static4)
     connect(seedToTime, &QAction::triggered, this, &Static4::seedToTime);
     ui->tableViewSearcher->addAction(seedToTime);
 
+    ui->comboBoxGeneratorShiny->setup({ toInt(Shiny::Never), toInt(Shiny::Random) });
+    ui->comboBoxSearcherShiny->setup({ toInt(Shiny::Never), toInt(Shiny::Random) });
+
     connect(ui->pushButtonGenerate, &QPushButton::clicked, this, &Static4::generate);
     connect(ui->pushButtonSearch, &QPushButton::clicked, this, &Static4::search);
     connect(ui->pushButtonProfileManager, &QPushButton::clicked, this, &Static4::profileManager);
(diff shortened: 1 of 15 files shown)


SteveCookTU:
Seeing as this is a static entry, I believe we should be following the same format as before and have it all contained within the templates rather than a check box just for a single Pokemon. If this means there are two static templates for locked vs non-locked then that should still be better.

Real96:
> Seeing as this is a static entry, I believe we should be following the same format as before and have it all contained within the templates rather than a check box just for a single Pokemon. If this means there are two static templates for locked vs non-locked then that should still be better.

Added the second template!

Source/Form/Gen4/Static4.ui
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/Gen4/Static4.cpp
```diff
@@ -330,6 +359,15 @@ void Static4::searcherPokemonIndexChanged(int index)
...
         }
+
+        if (staticTemplate->getShiny() == Shiny::Never)
+        {
+            ui->checkBoxSearcherShinyLock->setChecked(true);
+        }
+        else
+        {
+            ui->checkBoxSearcherShinyLock->setChecked(false);
+        }
```
Admiral-Fish, line 370: This should be able to be removed

Source/Form/Gen4/Static4.cpp
```diff
@@ -209,6 +220,13 @@ void Static4::generatorPokemonIndexChanged(int index)
             ui->labelGeneratorLead->setVisible(true);
             ui->comboMenuGeneratorLead->setVisible(true);
         }
+
+        if (staticTemplate->getShiny() == Shiny::Never) {
+            ui->checkBoxGeneratorShinyLock->setChecked(true);
+        }
+        else {
+            ui->checkBoxGeneratorShinyLock->setChecked(false);
+        }
```
Admiral-Fish, line 229: This should be able to be removed

Source/Form/Gen4/Static4.cpp
```diff
@@ -183,9 +183,20 @@ void Static4::generatorCategoryIndexChanged(int index)
...
+                        QString::fromStdString(Translator::getSpecie(templates[i].getSpecie(), templates[i].getForm()))
+                            + " (" + tr("Shiny Locked") + ")",
+                        QVariant::fromValue(i));
+                }
+                else
+                {
+                    ui->comboBoxGeneratorPokemon->addItem(
+                        QString::fromStdString(Translator::getSpecie(templates[i].getSpecie(), templates[i].getForm())),
+                        QVariant::fromValue(i));
+                }
```
Admiral-Fish, line 199: Do we want to do this or just add a shiny combobox that gets filled out by the template like we have in the other windows?

Source/Form/i18n/PokeFinder_de.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/i18n/PokeFinder_es.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/i18n/PokeFinder_fr.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/i18n/PokeFinder_it.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/i18n/PokeFinder_ja.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/i18n/PokeFinder_ko.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Source/Form/i18n/PokeFinder_zh.ts
Admiral-Fish, line 1: I don't believe this file needs to have any changes

Admiral-Fish (changes requested):

Source/Form/Gen4/Static4.cpp
Real96: Can't we have both? Or you prefer having only one of the two?

Imo, i would keep both. But if you want only one of the two, the shiny combobox would be better because "Shiny Locked" text is being cut.

Source/Form/Gen4/Static4.ui
Real96: Yes if we add the shiny combobox, right?

Real96: Added combobox instead of checkbox

Source/Form/Gen4/Static4.cpp
Real96: Done

Source/Form/Gen4/Static4.cpp
Real96: Done

Source/Form/Gen4/Static4.cpp
Real96: Done

Source/Form/i18n/PokeFinder_de.ts
Real96: Done

Source/Form/i18n/PokeFinder_es.ts
Real96: Done

Source/Form/i18n/PokeFinder_fr.ts
Real96: Done

Source/Form/i18n/PokeFinder_it.ts
Real96: Done

Source/Form/i18n/PokeFinder_ja.ts
Real96: Done

Source/Form/i18n/PokeFinder_ko.ts
Real96: Done

PR #430, opened 2024-10-17
Avoid compilation issue w/ Qt6 versions < 6.8

jkarns275:
Qt 6.8 is not available on MacOS via brew at the time so the code won't compile. I added a simple feature gate to use the old implementation of style. It compiles for me but I'm not 100% sure if it will actually function as intended since I don't know much about the themes in Qt or this codebase generally.


jkarns275:
References issue #431

jkarns275:
I ran into a few issues bundling the application as well... I can update the README w/ the things I had to do.

Issue #393, opened 2024-06-06
Flatpak


JakobDev:
I've created a [Flatpak](https://flatpak.org) package for PokeFinder, which allows easier Installation on Linux.
[Here are the build files](https://github.com/user-attachments/files/15605229/PokeFinder-linux.tar.gz). PokeFinder works (except 2 little patches) out of the box, so you just need to add some data files. I can make a PR which does all needed changes, if you are interested in an Flatpak package.

Admiral-Fish:
Is there any significant advantage to flatpak over something like appimage? I have see https://github.com/probonopd/linuxdeployqt before in the past but never tried to use it

Admiral-Fish:
What changes would the repo need to support that

Iguessthiswasinevitable:
I would also be in favor of a flatpak. They're a lot less dependent on external factors unique to your specific computer.

kaitouctr:
As a proof of concept, I have a repository (which is a fork of PokeFinder) that has the needed files and the correct `.yaml` workflow for `linuxdeploy` which creates AppImages. Like Flatpaks, they should also work the same across distributions, the difference is that distribution is similar to `.exe` or `.dmg`. I am not in favour of abandoning Flatpak, just in favour of expanding the means in which PokeFinder can be distributed on Linux, mGBA and melonDS are both distributed via both means.
[https://github.com/kaitouctr/PokeFinder/tree/appimage-gh-actions](https://github.com/kaitouctr/PokeFinder/tree/appimage-gh-actions)

kaitouctr:
By the way, for the Flatpak support, is there's any web domain that controlled by PokeFinder? That is particularly important for creating an application ID for Flatpak. I thought of `io.github._Admiral_Fish.PokeFinder`, which just requires Admiral-Fish to have a GitHub Pages set up. Those underscores are there as Flatpak (due to arbitrary dbus naming conventions) does not allow hyphens or dashes in anywhere but the last component of the application ID.

PR #440, opened 2024-11-11
Allow the Linux CI to produce AppImages

kaitouctr:
Adds the necessary files for the building of the AppImage (`.desktop` and the icon files) and makes changes to the CI workflow that allows it to produce AppImages via `linuxdeploy` (though the workflow retains its ability to produce a non-AppImage binary, and still uses its tar.gz for releases).

The AppStream, `.desktop` and icon files also allow the Flatpak request (described in #393) to be closer to fulfilled, with only the change to the `main.cpp` still needed to be done.


kaitouctr:
It's largely done under the principle of providing choice, but technically there is no need to keep the original binary unpackaged as the binary can still be built through following the build instructions or extracted from the AppImage by running the `--appimage-extract` option on it through Terminal.

kaitouctr:
Another small pedantic detail about this PR, it uses `linuxdeploy/linuxdeploy` rather than using `probonopd/linuxdeployqt` largely because `linuxdeployqt` requires building with a really old version of Ubuntu for some reason (apparently compatibility with really old systems).

PR #510, opened 2025-12-30
Fix shiny type for gen5 events

Real96:


Admiral-Fish (changes requested):
Leave the ui file alone. Instead add this to the cpp file

`
ui->comboBoxGeneratorShiny->setup({ toInt(Shiny::Never), toInt(Shiny::Random), toInt(Encounter::Always) });
`
`
ui->comboBoxSearcherShiny->setup({ toInt(Shiny::Never), toInt(Shiny::Random), toInt(Encounter::Always) });
`

PR #536, opened 2026-02-24
Split Emerald Egg in two tables

EliaValkyr:
In Emerald, eggs are created in two stages. Currently, pokefinder does a cartesian product of the two, leading to a crash when generating the table with the default settings, because it's trying to create 25 million rows. I have split the information in two tables, one for the held information (PID, nature, etc) and the other about the pickup information (IVs, hidden power) so that it's more convenient to use.

In my opinion, this also makes the tool easier to understand, since columns are now grouped by when they are populated in-game.

Here's how the table looks:

image

I tried to modify as little as possible from the code, hence why I've kept using the same state object for the two tables, `EggState3`. I've had to create new classes for the table model but those aren't very large so hopefully that's okay for you.

I've seen that for the rest of gen 3 games, something similar happens, as in the pokemon are created in two stages, though the information in each stage is different than in Emerald. If you like this change, I could also try and change those tables.

Anyway, please tell me what you think, and what changes you'd make to it!

### DIFF
--- a/Core/Gen3/Generators/EggGenerator3.cpp
+++ b/Core/Gen3/Generators/EggGenerator3.cpp
@@ -175,28 +175,10 @@ EggGenerator3::EggGenerator3(u32 initialAdvances, u32 maxAdvances, u32 offset, u
     }
 }
 
-std::vector<EggState3> EggGenerator3::generate(u32 seedHeld, u32 seedPickup) const
+std::vector<EggState3> EggGenerator3::generateRSFRLG(u32 seedHeld, u32 seedPickup) const
 {
-    switch (method)
-    {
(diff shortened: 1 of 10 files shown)


Admiral-Fish:
I believe PokeFinder used to be like this generating the results separately. I think as part of the redesign I did a couple years back I condensed to better utilize the information I had added (base stats, etc). I do think I had only a shared TableView between the two though that is a minor detail.

I'm not opposed to having a split it would just have the consequence of not being able to show correct stats as we wouldn't have nature. Perhaps an option to do separate or combined would be ideal.

@notblisy and @papajefe I think have complained to me about the new way before and can maybe provide extra insight.

Issue #563, opened 2026-04-28
[Gen 5 B2W2] Dream Radar Gen 4 legendaries are male in the results???


piplupness:
PokeFinder 4.3.2-da416e6 Windows

In #525 it appears they used to be genderless in the result table, but now all gen 4 legendaries are male???
It appears when Dialga, Palkia, Giratina, Ho-Oh, Lugia are selected, the table reports gender male despite you are not allowed to choose any since they are genderless.

Regular genderless Pokémon such as Staryu are not affected.

You can notice the ability Telepathy, Dialga:
<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/bdd988d0-344c-4be3-82b8-46ffcc39d80c" />

Ho-Oh (requested by @Real96):
<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/7d0b6618-b201-4887-9d3b-8c1eed3785d2" />

Lugia:
<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/5d9cd742-59ed-4dde-bc7c-5672fc1dad83" />

Staryu, appears normal:
<img width="1920" height="1020" alt="Image" src="https://github.com/user-attachments/assets/a06b1ba9-ad73-4fd3-a331-b2217bf88995" />

Admiral-Fish:
Fixed by #566

PR #566, opened 2026-05-03
Fix DreamRadar Gen4 legendaries wrong gender

Real96:
Fix for #563


Core/Gen5/Generators/DreamRadarGenerator.cpp
```diff
@@ -96,9 +96,10 @@ std::vector<DreamRadarState> DreamRadarGenerator::generate(u64 seed) const
 
         go.advance(2);
 
+        u8 gender = radarTemplate.getLegend() && !radarTemplate.getGenie() ? 2 : radarTemplate.getGender();
```
Admiral-Fish: This should be outside of the loop

Admiral-Fish (changes requested):

Core/Gen5/Generators/DreamRadarGenerator.cpp
Real96: Done

PR #575, opened 2026-06-07
[Gen 4 & 5] Addition of Chatot Pitch Searcher

NickPlayeZ:
This new feature lets you search for what RNG Advance you're currently on in gens 4 and 5 using Chatot pitches (or save needles in gen 5) in all the relevant generator tabs.

Also added a checkbox for gen 5 generators that lets you switch to save needles if you don't have chatot to advance the RNG.


Core/Gen5/Generators/EggGenerator5.cpp
```diff
@@ -174,7 +174,9 @@ std::vector<EggState5> EggGenerator5::generateBW(u64 seed) const
 
         u8 ability = hiddenAbility ? 2 : ((pid >> 16) & 1);
 
-        EggState5 state(rng.nextUInt(0x1fff), advances + initialAdvances + cnt, pid, ivs, ability, Utilities::getGender(pid, info), nature,
+        BWRNG rngCopy = rng;
+        u8 saveNeedle = static_cast<u8>(rngCopy.nextUInt(8));
+        EggState5 state(rng.nextUInt(0x1fff), saveNeedle, advances + initialAdvances + cnt, pid, ivs, ability, Utilities::getGender(pid, info), nature,
```
Admiral-Fish: Everywhere this is done I would prefer we just pass rng.nextUInt() a single time and let the constructor worry about doing a little extra math for chatot and needle

Same across all other files. I just won't leave a review comment on all of them.

Core/Util/Utilities.hpp
```diff
@@ -218,6 +218,19 @@ namespace Utilities5
...
+     * @param needle Save needle value
+     *
+     * @return Save needle arrow
+     */
+    inline std::string getSaveNeedle(u8 needle)
+    {
+        static const char *needles[] = { "↑", "↗", "→", "↘", "↓", "↙", "←", "↖" };
+        return needles[needle & 7];
+    }
+
```
Admiral-Fish: I usually put anything text related into Translator.hpp/Translator.cpp. I would split it with a declaration and a definition in those two files.

Form/Gen4/Eggs4.cpp
```diff
@@ -79,6 +83,19 @@ Eggs4::Eggs4(QWidget *parent) : QWidget(parent), ui(new Ui::Eggs4)
...
+    if (settingsLayout)
+    {
+        settingsLayout->addWidget(buttonAdvanceFinder, settingsLayout->rowCount(), 0, 1, 2);
+        connect(buttonAdvanceFinder, &QPushButton::clicked, this, &Eggs4::openAdvanceFinder);
+    }
+    else
+    {
+        delete buttonAdvanceFinder;
+    }
+
```
Admiral-Fish: Instead of a button this should probably be a context menu for the tableview. This file has some examples of adding QAction(s) to the table view

Same across all other files. I just won't leave a review comment on all of them.

Form/Gen4/Eggs4.hpp
```diff
@@ -66,6 +67,7 @@ public slots:
 private:
     Ui::Eggs4 *ui;
 
+    AdvanceFinder *advanceFinder;
```
Admiral-Fish: This doesn't need to be part of the private data. Should probably follow the model I use for the profile manager.

Same across all other files. I just won't leave a review comment on all of them.

Model/Gen5/EggModel5.hpp
```diff
@@ -83,12 +83,20 @@ public slots:
      */
     void setShowStats(bool flag);
 
+    /**
```
Admiral-Fish: Might make more sense to just not have this be toggleable and just show needles next to the chatot.

Same across all other files. I just won't leave a review comment on all of them.

Form/CMakeLists.txt
```diff
@@ -176,6 +176,8 @@ set(PROJECT_SOURCES
     Gen5/DreamRadar.hpp
     Gen5/Eggs5.cpp
     Gen5/Eggs5.hpp
+    Gen5/AdvanceFinder.cpp
+    Gen5/AdvanceFinder.hpp
```
Admiral-Fish: I normally put things that are utilized by multiple gens to Form/Util

Form/Gen5/Eggs5.ui
```diff
@@ -260,6 +260,13 @@
           <item row="0" column="0">
            <widget class="EggSettings" name="eggSettingsGenerator" native="true"/>
           </item>
+          <item row="1" column="0" colspan="2">
+           <widget class="QCheckBox" name="checkBoxGeneratorSaveNeedles">
+            <property name="text">
+             <string>Display Save Needles</string>
+            </property>
+           </widget>
+          </item>
```
Admiral-Fish: This can be removed if we just always show the needle.

Same for the other .ui files

Form/Gen5/Eggs5.cpp
```diff
@@ -65,6 +69,32 @@ Eggs5::Eggs5(QWidget *parent) : QWidget(parent), ui(new Ui::Eggs5)
...
+        layout->addWidget(ui->checkBoxGeneratorSaveNeedles);
+        layout->addWidget(buttonAdvanceFinder);
+        layout->addStretch();
+        parentLayout->addLayout(layout, row, column, rowSpan, columnSpan);
+        connect(buttonAdvanceFinder, &QPushButton::clicked, this, &Eggs5::openAdvanceFinder);
+    }
+    else
+    {
+        delete buttonAdvanceFinder;
+    }
```
Admiral-Fish: Can't say I fully follow what is happening here but with the changes that would remove the need for the checkbox and moving the Advance Finder to the tableview context menu I don't believe most of this will be necessary.

Same for the other files.

Form/Gen5/AdvanceFinder.hpp
```diff
@@ -0,0 +1,176 @@
...
+    ValueGetter chatotGetter;
+    ValueGetter needleGetter;
+    QRadioButton *radioButtonChatot;
+    QRadioButton *radioButtonNeedle;
+    QGroupBox *groupBoxChatot;
+    QGroupBox *groupBoxNeedle;
+    QLineEdit *lineEditSequence;
+    QLabel *labelPossibleResults;
+    QPushButton *buttonJump;
+    QTableWidget *tableWidgetResults;
```
Admiral-Fish: I use the .ui file model and use Qt Creator or Qt Widgets Designer to layout the UI, you should create one for this.

Form/Gen5/AdvanceFinder.cpp
```diff
@@ -0,0 +1,537 @@
...
+    QStringList headers;
+    for (int i = 0; i < generatorModel->columnCount(); i++)
+    {
+        headers << generatorModel->headerData(i, Qt::Horizontal, Qt::DisplayRole).toString();
+    }
+    tableWidgetResults->setHorizontalHeaderLabels(headers);
+    tableWidgetResults->setEditTriggers(QAbstractItemView::NoEditTriggers);
+    tableWidgetResults->setSelectionBehavior(QAbstractItemView::SelectRows);
+    tableWidgetResults->horizontalHeader()->setSectionResizeMode(QHeaderView::Interactive);
+    mainLayout->addWidget(tableWidgetResults);
```
Admiral-Fish: Most of this will end up in the .ui file

Form/Gen5/AdvanceFinder.cpp
```diff
@@ -0,0 +1,537 @@
...
+
+        if (match)
+        {
+            matches.emplace_back(i);
+        }
+    }
+
+    labelPossibleResults->setText(tr("Possible Results: %1").arg(matches.size()));
+    displayResults(matches, sequence.size());
+}
```
Admiral-Fish: I generally keep "backend" logic in Core instead of Form

Form/Gen5/AdvanceFinder.hpp
```diff
@@ -0,0 +1,176 @@
...
+    AdvanceFinder(EventGeneratorModel4 *generatorModel, QWidget *parent = nullptr);
+    AdvanceFinder(EventGeneratorModel4 *generatorModel, QTableView *sourceTableView, QWidget *parent = nullptr);
+    AdvanceFinder(EggGeneratorModel5 *generatorModel, QWidget *parent = nullptr);
+    AdvanceFinder(EggGeneratorModel5 *generatorModel, QTableView *sourceTableView, QWidget *parent = nullptr);
+    AdvanceFinder(EggGeneratorModel4 *generatorModel, QWidget *parent = nullptr);
+    AdvanceFinder(EggGeneratorModel4 *generatorModel, QTableView *sourceTableView, QWidget *parent = nullptr);
+    AdvanceFinder(HiddenGrottoSlotGeneratorModel5 *generatorModel, QWidget *parent = nullptr);
+    AdvanceFinder(HiddenGrottoSlotGeneratorModel5 *generatorModel, QTableView *sourceTableView, QWidget *parent = nullptr);
+    AdvanceFinder(HiddenGrottoGeneratorModel5 *generatorModel, QWidget *parent = nullptr);
+    AdvanceFinder(HiddenGrottoGeneratorModel5 *generatorModel, QTableView *sourceTableView, QWidget *parent = nullptr);
```
Admiral-Fish: This seems like a bit much. We could probably just pass the parent QAbstractTableModel and use model->data() to get what we need.

Form/Gen5/AdvanceFinder.cpp
```diff
@@ -0,0 +1,537 @@
...
+            sequence.emplace_back(chatotRange(40, 79));
+        }
+        else if (token == "MID/MID-LOW" || token == "MID/MIDLOW" || token == "M/ML")
+        {
+            sequence.emplace_back(chatotRange(20, 59));
+        }
+        else if (token == "MID-LOW/LOW" || token == "MIDLOW/LOW" || token == "ML/L")
+        {
+            sequence.emplace_back(chatotRange(0, 39));
+        }
```
Admiral-Fish: All of these checks seem like overkill. We only need to check against the text that is being added from the buttons. Also they all need to be wrapped in tr() calls otherwise it won't work in other languages.

The line input should be read-only if it isn't already.

Form/Gen5/AdvanceFinder.cpp
```diff
@@ -0,0 +1,537 @@
...
+{
+    std::vector<QVector<u8>> sequence;
+    valid = true;
+
+    QStringList tokens = lineEditSequence->text().split(',', Qt::SkipEmptyParts);
+    for (QString token : tokens)
+    {
+        token = token.trimmed().toUpper();
+
+        if (token == "ANY" || token == "A" || token == "*")
```
Admiral-Fish: Same comment as above (R340-R379).

Also does it even make sense to have an "Any" for needles. They are very easy to tell what they are. I could understand having it for chatot if you have trouble identifying pitch but I don't see a similar argument for needles.

Form/Gen5/AdvanceFinder.cpp
```diff
@@ -0,0 +1,537 @@
...
+    }
+    return values;
+}
+
+static QString needleSymbol(u8 needle)
+{
+    static const char *needles[] = { "\342\206\221", "\342\206\227", "\342\206\222", "\342\206\230",
+                                     "\342\206\223", "\342\206\231", "\342\206\220", "\342\206\226" };
+    return QString::fromUtf8(needles[needle & 7]);
+}
```
Admiral-Fish: Can just get the needle text from Translator once you make that change and then QString::fromStdString()

Form/Gen5/AdvanceFinder.cpp
```diff
@@ -0,0 +1,537 @@
...
+static QVector<u8> chatotRange(u8 min, u8 max)
+{
+    QVector<u8> values;
+    values.reserve(max - min + 1);
+    for (u8 value = min; value <= max; value++)
+    {
+        values.append(value);
+    }
+    return values;
+}
```
Admiral-Fish: Might be simpler to do a tuple here instead of constructing a whole vector. Then when used below would just be

```
for (i = range.first; i < range.second; i++)
```

Admiral-Fish (changes requested):

PR #576, opened 2026-06-15
[Gen 5] Adjacent seed tool

NickPlayeZ:
This PR adds Adjacent Seed Tool to the gen 5 tools (and you can right click into a gen 5 searcher tab for wild, static or grotto rng to open it with the seeds date/time/keypresses).


Form/Controls/TableView.cpp
```diff
@@ -61,6 +61,12 @@ void TableView::contextMenuEvent(QContextMenuEvent *event)
 {
     if (model()->rowCount() != 0)
     {
+        QModelIndex index = indexAt(event->pos());
+        if (index.isValid())
+        {
+            selectRow(index.row());
+            setCurrentIndex(index);
+        }
```
Admiral-Fish, line 69: What does this accomplish?

NickPlayeZ: it makes sure that by right clicking on a specific row in the searcher tabs results, adjacent seed tool opens with that date/time/keypress combo. Otherwise if the user had selected a different row by clicking into it beforehand, that would be the one that adjacent seed tool would take the date/time/keypresses from instead

Core/Gen5/Tools/AdjacentSeedsCalculator.cpp
```diff
@@ -0,0 +1,133 @@
...
+                sha.setTimer0(timer0, profile.getVCount());
+                auto alpha = sha.precompute();
+                u64 seed = sha.hashSeed(alpha);
+
+                for (u32 ivAdvance = minIVAdvance; ivAdvance <= maxIVAdvance; ivAdvance++)
+                {
+                    std::array<u8, 6> ivs = generateMTIVs(seed, profile.getVersion(), ivAdvance, roamer);
+
+                    states.emplace_back(seed, offset, buttons, static_cast<u16>(timer0), ivAdvance, ivs,
+                                        Utilities5::initialAdvances(seed, profile) + ivAdvance,
```
Admiral-Fish: This feels wrong. Utilities5::initialAdvances gets us initial PID advances and you are adding IV advances to it. What is the intention here?

Form/Gen5/Tools/AdjacentSeeds.cpp
```diff
@@ -0,0 +1,311 @@
...
+    model = new AdjacentSeedsModel(ui->tableView);
+    ui->tableView->setModel(model);
+    ui->tableView->setItemDelegate(new TargetRowDelegate(ui->tableView));
+
+    keypressModel = new QStandardItemModel(this);
+    ui->comboBoxKeypresses->setModel(keypressModel);
+    ui->comboBoxKeypresses->setEditable(true);
+    ui->comboBoxKeypresses->lineEdit()->setReadOnly(true);
+    ui->comboBoxKeypresses->lineEdit()->installEventFilter(this);
+    ui->comboBoxKeypresses->setMaxVisibleItems(static_cast<int>(keypressButtons.size()));
```
Admiral-Fish: I had at some point intended to try to make CheckList behave properly for this. Not sure if I gave up or not, looking at ProfileCalibrator5 I used a scrollarea with a number of checkboxes. Maybe I will spend some more time seeing if I can get CheckList to behave or maybe just a new form class to provide this type of functionality that can be reused.

Admiral-Fish (changes requested):

Core/Gen5/Tools/AdjacentSeedsCalculator.cpp
NickPlayeZ: Yeah its wrong; overlooked error. Intention of this part is just to generate chatters and save needles at the bottom. Should be Utilities5::initialAdvances(seed, profile) instead

Form/Gen5/Tools/AdjacentSeeds.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    void initStyleOption(QStyleOptionViewItem *option, const QModelIndex &index) const override
+    {
+        QStyledItemDelegate::initStyleOption(option, index);
+        if (index.data(Qt::UserRole).toBool())
+        {
+            option->font.setBold(true);
+        }
+    }
+};
```
Admiral-Fish, line 52: What exactly is this accomplishing?

Admiral-Fish (commented):
I think I have done all the changes I wanted to do. Let me know if you disagree or have any questions

Form/Gen5/Tools/AdjacentSeeds.cpp
NickPlayeZ: It bolds the row that has the combination of exact target time + timer0 min in the Adjacent Seed Tool results table

Its primarily to make it easier to identify at a glance what the target seed is and rng reporter had this too. It's less useful for bw2 as the timer0 min of a larger timer0 range is rarely the target timer0, but for bw1 its usually a little bit useful

NickPlayeZ:
> I think I have done all the changes I wanted to do. Let me know if you disagree or have any questions

there is one thing thats super niche but has some use to keep; if your target time is the 31st of december 2099 at 23:59 and like 59 seconds and you overshoot by 1 second, the DS loops back to the year 2000. So keeping the loop in adjacent seed tool would be useful there.
Now it is possible that this year rollover messes with the rng in some way, i didnt test that, but just in case it works "smoothly", id keep it displaying 2000 as the year after 2099

Admiral-Fish:
> > I think I have done all the changes I wanted to do. Let me know if you disagree or have any questions
> 
> there is one thing thats super niche but has some use to keep; if your target time is the 31st of december 2099 at 23:59 and like 59 seconds and you overshoot by 1 second, the DS loops back to the year 2000. So keeping the loop in adjacent seed tool would be useful there. Now it is possible that this year rollover messes with the rng in some way, i didnt test that, but just in case it works "smoothly", id keep it displaying 2000 as the year after 2099

Is that even a common thing to do? I usually just RNG between 2000 and the current year. Would it be confusing for a user to see that? I didn't even know the date rolls back over

NickPlayeZ:
> > > I think I have done all the changes I wanted to do. Let me know if you disagree or have any questions
> 
> > 
> 
> > there is one thing thats super niche but has some use to keep; if your target time is the 31st of december 2099 at 23:59 and like 59 seconds and you overshoot by 1 second, the DS loops back to the year 2000. So keeping the loop in adjacent seed tool would be useful there. Now it is possible that this year rollover messes with the rng in some way, i didnt test that, but just in case it works "smoothly", id keep it displaying 2000 as the year after 2099
> 
> 
> 
> Is that even a common thing to do? I usually just RNG between 2000 and the current year. Would it be confusing for a user to see that? I didn't even know the date rolls back over

It would definitely be extremely rare, but can happen. And I feel it'd be more confusing for a user that does rng with that specific date/time to just not be able to view the adjacent seeds on a higher target second when they do exist

Admiral-Fish (approved):

PR #578, opened 2026-06-17
show hidden advances for fishing, rock smash and grottoes (gens 3, 4 and 5)

NickPlayeZ:


Model/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 project(PokeFinderModel)
 
+find_package(Qt6 REQUIRED COMPONENTS Core Gui)
```
Admiral-Fish: Just add Gui to the top level CMakeLists.txt file. It gets inherited down.

Model/Gen5/WildModel5.cpp
```diff
@@ -32,10 +34,29 @@ int WildGeneratorModel5::columnCount(const QModelIndex &parent) const
...
+
     if (role == Qt::DisplayRole)
     {
-        const auto &state = model[index.row()];
         int column = index.column();
+        if (!state.isValid() && column > 1)
+        {
+            return "-";
+        }
+
```
Admiral-Fish: Should we just show the original data or does it make more sense to show dashes?

Same for all other Model files

Core/Gen4/Generators/WildGenerator4.hpp
```diff
@@ -67,6 +68,7 @@ class WildGenerator4 : public WildGenerator<EncounterArea4, Profile4, WildStateF
     bool shiny;
     bool unownRadio;
     u8 happiness;
+    bool includeEmpty;
```
Admiral-Fish: What purpose does this variable serve? Would it not just make more sense to always add the entries?

Same for the other files

Admiral-Fish (changes requested):
Two general notes:

1. Gen 3 has rock smash that is missing from this
2. Might make sense to move the valid flag to WildGeneratorState so Gen 3/4/5 can all use it

Model/Gen5/WildModel5.cpp
NickPlayeZ: I think in most cases it makes sense to just omit the original data entirely as to not confuse users so they think they may be able to hit that feasibly; unless there are always methods to increase the amount of advances you can feasibly get an encounter by a noteworthy lot. Not sure myself

Admiral-Fish:
I moved the hiding into filter. Let me know if you have any concerns

NickPlayeZ:
> I moved the hiding into filter. Let me know if you have any concerns

Generally a useful change, but I think it should be disabled by default; 99% of the time people will want to track the advances via e.g. chatot (at least in gens 4 and 5).

Also I feel like its confusing to the average joe to have the option no matter what wild encounter type is selected considering it only has an effect for a specific set of encounter types like rock smash or fishing

(also gen 3 rock smash currently does not have this box; is that intentional? I guess thinking about it there's no benefit there since theres no chatot unless I'm missing something)

Form/Gen4/Wild4.cpp
```diff
@@ -417,6 +417,8 @@ void Wild4::generatorEncounterIndexChanged(int index)
             || encounter == Encounter::GoodRod || encounter == Encounter::SuperRod;
         bool honey = encounter == Encounter::HoneyTree;
 
+        ui->filterGenerator->setHideInvalidVisible(fish || (hgss && encounter == Encounter::RockSmash));
```
Admiral-Fish: I don't think we technically need the hgss check.

Admiral-Fish (changes requested):
I removed the option for Gen 3 rock smash as there is no useful data like chatot to be referencing.

NickPlayeZ:
> I removed the option for Gen 3 rock smash as there is no useful data like chatot to be referencing.

Makes sense

Admiral-Fish (approved):
I believe this should be good to merge if you are happy with everything

PR #581, opened 2026-06-24
[Gen 5] Addition of Swarms

NickPlayeZ:
Adds Swarm encounters to Gen 5 wild RNG.


Admiral-Fish (commented):
I was thinking of handling this like how I do Feebas tile. It is just the next available slot: so in this case 12.

Would simplify a lot of this PR, probably wouldn't need any of the changes in EncounterArea5.hpp. EncounterArea.hpp would need a very small change to increase array size which would likely slightly spiral to all the Encounters*.cpp files to increase the array size.

I don't think it's super important to specially show it as slot "Swarm" vs slot "12". I don't give feebas any special displaying

NickPlayeZ:
I do still think it'd be nicer to have the slot be called swarms just to make it clearer to users that this is slot does not abide by the regular encounter slot logic

NickPlayeZ:
if you're extremely against, we can always roll back the two commits, but I do believe that this is just a nice and small improvement

Core/Gen5/EncounterArea5.hpp
```diff
@@ -66,14 +66,15 @@ class EncounterArea5 : public EncounterArea
...
+                if (other.getMinLevel() != other.getMaxLevel())
                 {
                     diff = true;
                 }
 
-                if (slot.getSpecie() == pokemon[i].getSpecie())
+                if (slot.getSpecie() == other.getSpecie())
                 {
-                    max = std::max(max, pokemon[i].getMaxLevel());
+                    max = std::max(max, other.getMaxLevel());
```
Admiral-Fish: This didn't need to be changed

Core/Parents/Filters/StateFilter.cpp
```diff
@@ -183,14 +184,23 @@ bool StateFilter::compareState(const State8 &state) const
...
+                                 const std::array<bool, 16> &powers, const std::array<bool, 13> &encounterSlots) :
     StateFilter(gender, ability, shiny, levelMin, levelMax, heightMin, heightMax, weightMin, weightMax, skip, ivMin, ivMax, natures, powers),
     encounterSlots(encounterSlots)
 {
 }
 
 bool WildStateFilter::compareEncounterSlot(u8 encounterSlot) const
 {
-    return skip || encounterSlots[encounterSlot];
+    return skip || (encounterSlot < encounterSlots.size() && encounterSlots[encounterSlot]);
```
Admiral-Fish: What do these changes accomplish? encounterSlots probably just needs to be changed to array of size 13

Core/Parents/Filters/StateFilter.hpp
```diff
@@ -198,6 +198,28 @@ class WildStateFilter : public StateFilter
...
+     * @param skip If filters should be skipped
+     * @param min Minimum IV thresholds
+     * @param max Maximum IV thresholds
+     * @param natures Natures to filter by
+     * @param powers Hidden powers to filter by
+     */
+    WildStateFilter(u8 gender, u8 ability, u8 shiny, u8 levelMin, u8 levelMax, u8 heightMin, u8 heightMax, u8 weightMin, u8 weightMax, bool skip,
+                    const std::array<u8, 6> &ivMin, const std::array<u8, 6> &ivMax, const std::array<bool, 25> &natures,
+                    const std::array<bool, 16> &powers, const std::array<bool, 13> &encounterSlots);
+
```
Admiral-Fish, line 208: An extra constructor seems unnecessary. Just increase the array size in the other one

Form/Controls/Filter.hpp
```diff
@@ -109,6 +109,13 @@ class Filter : public QWidget
      */
     std::array<bool, 12> getEncounterSlots() const;
 
+    /**
+     * @brief Gets encounter slots to filter by for wild encounters
+     *
+     * @return Array of encounter slots
+     */
+    std::array<bool, 13> getWildEncounterSlots() const;
+
```
Admiral-Fish: Just make getEncounterSlots() return array of 13

Form/Controls/Filter.cpp
```diff
@@ -283,6 +284,14 @@ std::array<bool, 12> Filter::getEncounterSlots() const
...
 }
 
+std::array<bool, 13> Filter::getWildEncounterSlots() const
+{
+    auto checked = ui->checkListEncounterSlot->getChecked();
+    std::array<bool, 13> array = {};
+    std::copy_n(checked.begin(), std::min(checked.size(), array.size()), array.begin());
+    return array;
+}
+
```
Admiral-Fish: Just make getEncounterSlots() return array of 13

ui->checkListEncounterSlot->getCheckedArray<13>();

Form/Gen5/Wild5.cpp
```diff
@@ -227,7 +229,12 @@ void Wild5::generate()
     auto lead = ui->comboMenuGeneratorLead->getEnum<Lead>();
     u8 luckyPower = ui->comboBoxGeneratorLuckyPower->getCurrentUChar();
 
-    auto filter = ui->filterGenerator->getFilter<WildStateFilter, true>();
+    WildStateFilter filter(ui->filterGenerator->getGender(), ui->filterGenerator->getAbility(), ui->filterGenerator->getShiny(),
+                           ui->filterGenerator->getLevelMin(), ui->filterGenerator->getLevelMax(), ui->filterGenerator->getHeightMin(),
+                           ui->filterGenerator->getHeightMax(), ui->filterGenerator->getWeightMin(), ui->filterGenerator->getWeightMax(),
+                           ui->filterGenerator->getDisableFilters(), ui->filterGenerator->getMinIVs(), ui->filterGenerator->getMaxIVs(),
+                           ui->filterGenerator->getNatures(), ui->filterGenerator->getHiddenPowers(),
+                           ui->filterGenerator->getWildEncounterSlots());
```
Admiral-Fish: This should be able to be reverted after other changes

Form/Gen5/Wild5.cpp
```diff
@@ -395,7 +412,12 @@ void Wild5::search()
     auto lead = ui->comboMenuSearcherLead->getEnum<Lead>();
     u8 luckyPower = ui->comboBoxSearcherLuckyPower->getCurrentUChar();
 
-    auto filter = ui->filterSearcher->getFilter<WildStateFilter, true>();
```
Admiral-Fish, line 408: This should be able to be reverted after other changes

Model/Gen5/WildModel5.cpp
```diff
@@ -48,7 +48,7 @@ QVariant WildGeneratorModel5::data(const QModelIndex &index, int role) const
             return QString::fromStdString(Translator::getItem(state.getItem()));
         case 4:
             return QString("%1: %2")
-                .arg(state.getEncounterSlot())
```
Admiral-Fish, line 72: Is "S" really sufficient? I'm still not sure how I feel about needing to display it separately then 12

Admiral-Fish (changes requested):

Core/Gen5/EncounterArea5.hpp
NickPlayeZ: fixed

Core/Parents/Filters/StateFilter.cpp
NickPlayeZ: changed to 13

Core/Parents/Filters/StateFilter.hpp
NickPlayeZ: fixed

Form/Controls/Filter.hpp
NickPlayeZ: fixed

Form/Controls/Filter.cpp
NickPlayeZ: fixed

Form/Gen5/Wild5.cpp
NickPlayeZ: fixed

Form/Gen5/Wild5.cpp
NickPlayeZ: fixed

Model/Gen5/WildModel5.cpp
NickPlayeZ: I just think it makes it a bit clearer that this isnt simply another encounter slot, but works a bit differently. Not a big improvement though

Admiral-Fish: I'll still not sure how I feel about this. I don't do anything special for Feebas tiles and those work differently. Seems like it would be breaking consistency.

NickPlayeZ: Perhaps you should do something special for feebas then 😎
Encounter slot F or something idk

Admiral-Fish: I gave feebas as the most related example. But even in gen 4 nothing special is shown for all the encounter slot replacement going in. I am going to revert this

Admiral-Fish (approved):
This should be good to go now I believe. I moved the swarm encounter data to the EncounterTables submodule and fixed the proc check for swarm.

Model/Gen5/WildModel5.cpp
NickPlayeZ: Alright then

Issue #573, opened 2026-05-19
[Gen 5] return error message when trying to breed for Hidden Ability with impossible parent combination


NickPlayeZ:
Right now if you have e.g. a ditto and an HA pokemon as parents, it just finds nothing. I've seen at least a dozen people ask why they couldnt find anything over the past 2 years.

It would be cool if it could return an error message explaining "female parent must have HA and other parent cannot be ditto" or something whenever the parent combo cannot yield HA (e.g. male ha parent + female regular ability parent, ditto + any ha parent or 2 regular ability parents).

Issue #460, opened 2025-06-02
Have IV calculator game selection also change characteristic names


NickPlayeZ:
2 of the HP characteristics have been renamed in gen 6 and have those gen 6+ names in the PokeFinder IV calculator no matter what. Can this be changed so that if you select a gen 3, 4 or 5 game, the characteristics dropdown has their older gen names?
![Image](https://github.com/user-attachments/assets/5a2053a1-1fe8-4db1-975b-9611179430bf)

NickPlayeZ:
> How important is this in practice? I made a decision to use the newer names in [#221](https://github.com/Admiral-Fish/PokeFinder/issues/221)

It does confuse people from time to time (literally had such a case today in the pokemonrng discord lol) and since, at least at the moment, pokefinder is largely used for gens 3-5, I think it would make sense to at the very least include the old names when switching the IV calc to those games.

![Image](https://github.com/user-attachments/assets/8dec9062-96aa-4ba6-924c-6d750662f163)

Admiral-Fish:
Does this only impact English text or the other languages as well?

NickPlayeZ:
(sry for lare reply didnt check github for a while)
There are differences in other languages as well yes.

Like in german for example the spdef characteristic for 1, 6, 11, 16, 21, 26, 31 spdef was "Irgendwie Eitel" in gens 4 and 5, "Etwas Eitel" in gen 6 and "Ist etwas Eitel" in gens 6-8 which are all easily similar enough to know which is supposed to be it.

But then there's others like the defense characteristic for 2, 7, 12, 17, 22, 27  defense which was "Äußerst Ausdauernd" in gens 4-6, "Ist äußerst ausdauernt" in gen 7 and then got changed to "Ist äußerst zäh" in gen 8+. Most natives could make the connection there but its not nearly as obvious as the first example.

Admiral-Fish:
I wouldn't mind supporting this, I put together a prototype branch. Do you know if there is a complete list of these differences anywhere?

NickPlayeZ:
I wouldn't know of any unfortunately. Bulbapedia ofc has the english ones though with the german ones at the bottom of the page seem to not be fully updated for gens 8+. Pokewiki has the german ones here https://www.pokewiki.de/Individuelle_St%C3%A4rken#Pers%C3%B6nlichkeiten
[...]

PR #585, opened 2026-07-06
Quality of Life changes

NickPlayeZ:
-fixing visual bug that made pokemon names sometimes have () behind them

-expand ComboBox autofill to also suggest autofills if the user inputs later parts of the word (e.g. alph for ruins of alph)

-add delay display to gen 4 searcher tabs so the user can much faster decide what target they wish to go for without needing to look up each delay on seed to time

-add hour display to gen wild and static searchers for time based wild encounters + rotom, so the user doesnt have to open an extra window (seed to time) just to check what results are valid

-change gen 4 wild default time to day since its the largest time window thus making it most likely to be the currently relevant one on average

-expand search calls/flips windows in gen 4 seed to time to display the results inside the window already (3 or less results), so a user doesnt always have to close and re-open the search calls/flips window

-fix error in seed to time with column header names for hgss calls and roamers

-return error message when trying to breed for Hidden Ability with impossible parent combination #573 

-add generation based characteristic names #460 (data from the Poke Corpus database https://abcboy101.github.io/poke-corpus/#)


Admiral-Fish:
Separate unrelated changes into multiple PRs.

PR #586, opened 2026-07-06
[Gen 4] Expand search calls/flips windows in seed to time + fix error with column header names

NickPlayeZ:
-expand search calls/flips windows in gen 4 seed to time to display the results inside the window already (3 or less results), so a user doesnt always have to close and re-open the search calls/flips window

-fix error in seed to time with column header names for hgss calls and roamers


Admiral-Fish:
What exactly is the use case that makes this useful? Not saying I disagree with the idea just curious how useful this actually is

NickPlayeZ:
> What exactly is the use case that makes this useful? Not saying I disagree with the idea just curious how useful this actually is

Ive very often wanted to re-check what coin flips or calls I input after I made it mark me a result. I can't imagine I'm the only one either. This way the user doesn't have to close the flips/calls window every time and can view the results inside the small window to not need to switch back to the main seed to time window.
Not a major problem, but certainly a nice and small QOL improvement in my opinion

Admiral-Fish (commented):
Let me know if you disagree with anything I changed or need me to explain anything. Otherwise I think this is good to merge.

NickPlayeZ:
> Let me know if you disagree with anything I changed or need me to explain anything. Otherwise I think this is good to merge.

Will check tomorrow

NickPlayeZ:
> Let me know if you disagree with anything I changed or need me to explain anything. Otherwise I think this is good to merge.

generally looks good to me. Just expanded the date/time column in the search flips/calls window so the date and time are actually fully shown; especially important for the second hit.

Before
grafik

After
grafik

Admiral-Fish:
I changed the tableviews to use my custom TableView class which handles the resizing stuff

NickPlayeZ:
> I changed the tableviews to use my custom TableView class which handles the resizing stuff

this is perfect now!

Admiral-Fish (approved):

PR #588, opened 2026-07-07
Addition of Generation based characteristic names

NickPlayeZ:
-adds generation based characteristic names, so that theyre not wrong anymore for pre gen 6. #460 (data from the Poke Corpus database https://abcboy101.github.io/poke-corpus/#)


Form/Util/IVCalculator.cpp
```diff
@@ -229,6 +229,21 @@ void IVCalculator::gameIndexChanged(int index)
...
+        if (hasCharacteristics)
+        {
+            int index = ui->comboBoxCharacteristic->currentIndex();
+            ui->comboBoxCharacteristic->clear();
+            ui->comboBoxCharacteristic->addItem(tr("None"));
+            ui->comboBoxCharacteristic->addItems(Translator::getCharacteristics(version));
+            if (index != -1)
+            {
+                ui->comboBoxCharacteristic->setCurrentIndex(index);
+            }
```
Admiral-Fish, line 243: Is it even useful to persist the index?

NickPlayeZ: a little bit; and I dont see any hurt in having it work this way

sometimes i input stats before inputting the species and if I'm still at the base gen 3 games but my species is a gen 4 or gen 5 mon, it's neat if the iv calc keeps the characteristic

it also does save the char in the current main pokefinder if the user switches between games

Admiral-Fish: The main reason it currently saves it is because the field never gets cleared. We can maintain behavior with this

Regardless if you are okay with the changes I made this can merge

NickPlayeZ: Didn't get to checking them yet; will do so tomorrow

NickPlayeZ:
all looks good! Merge away :)

Admiral-Fish (approved):

PR #589, opened 2026-07-07
[Gen 5] Return error message when trying to breed for Hidden Ability with impossible parent combination 

NickPlayeZ:
-return error message when trying to breed for Hidden Ability with impossible parent combination #573


Admiral-Fish (commented):
I think it will probably be better long term to have a isValid() function like the Filter class has. EggSettings technically already has compatibleParents() function which could be repurposed and could move the gender incompatibility check message in there as well.

What do you think about that?

NickPlayeZ:
> I think it will probably be better long term to have a isValid() function like the Filter class has. EggSettings technically already has compatibleParents() function which could be repurposed and could move the gender incompatibility check message in there as well.
> 
> What do you think about that?

I do not care how this is handled on the codes side; just the functionality to the user matters to me so I'm on board with that lol

Form/Controls/EggSettings.cpp
```diff
@@ -125,6 +125,36 @@ bool EggSettings::compatibleParents() const
...
+    u8 ability1 = ui->comboBoxParentAAbility->currentIndex();
+    u8 ability2 = ui->comboBoxParentBAbility->currentIndex();
+
+    bool hiddenAbilityCompatible = (parent1 == 0 && parent2 == 1 && ability2 == 2) || (parent1 == 1 && ability1 == 2 && parent2 == 0);
+    if (hiddenAbility && !hiddenAbilityCompatible)
+    {
+        QMessageBox box(QMessageBox::Warning, tr("Incompatible Parents"), tr("Parents incompatible for breeding Hidden Ability!"));
+        box.exec();
+        return false;
+    }
```
Admiral-Fish: Does this work for gen 8?

NickPlayeZ: It's not currently used for gen 8; I know hidden ability breeding has other rules for gen 8, but I dont know how it works there off the top of my head. Should I check that out and implement this for gen 8 as well? I just really dont know shit about gen 8 rng xD

Admiral-Fish: May as well be complete with the change

NickPlayeZ: I'll get on it tomorrow

NickPlayeZ: done

Admiral-Fish (approved):

PR #593, opened 2026-07-11
[Gen 5] Pickup RNG implementation

NickPlayeZ:
Adds full Gen 5 Pickup RNG support to Pokefinder


Admiral-Fish:
This UI setup is a mess, doesn't even conform slightly to how I do it anywhere else. That will have to change before I look at this

NickPlayeZ:
> This UI setup is a mess, doesn't even conform slightly to how I do it anywhere else. That will have to change before I look at this

Will look into it tomorrow if there's still time after the other stuff or latest Monday

NickPlayeZ:
Should be fine now

NickPlayeZ:
accident happened without me realizing; resolved now

Admiral-Fish:
Why does this basically reimplement Wild RNG? It just feels wrong and I don't like that, why aren't we just displaying only the items.

Core/Gen5/Generators/PickupGenerator.hpp
```diff
@@ -0,0 +1,77 @@
...
+    };
+
+    PickupGenerator(u32 initialAdvances, u32 maxAdvances, const std::array<Slot, 6> &pickupSlots, const Profile5 &profile,
+                    bool includeInvalid = false);
+    PickupGenerator(u32 initialAdvances, u32 maxAdvances, const std::array<Slot, 6> &pickupSlots, const EncounterArea5 &area,
+                    const Profile5 &profile, bool includeInvalid = false);
+    PickupGenerator(u32 initialAdvances, u32 maxAdvances, const std::array<Slot, 6> &pickupSlots, const std::vector<Request> &requests,
+                    const EncounterArea5 &area, const Profile5 &profile, bool includeInvalid = false);
+    PickupGenerator(u32 initialAdvances, u32 maxAdvances, const std::array<Slot, 6> &pickupSlots, const EncounterArea5 &area,
+                    const Profile5 &profile, const WildStateFilter &filter, bool includeInvalid = false);
```
Admiral-Fish: Why do we have 4 different constructors? I can't currently think of any other Generator class that has this many constructors.

Form/Gen5/Pickup.hpp
```diff
@@ -0,0 +1,110 @@
...
+    std::array<ComboBox *, 6> searcherItemCounts;
+    std::array<QPushButton *, 6> searcherItemClearButtons;
+    TextBox *generatorIVAdvances;
+    ComboBox *generatorEncounter;
+    ComboBox *generatorSeason;
+    ComboBoxProxy *generatorLocation;
+    ComboBox *generatorPokemon;
+    ComboBox *searcherEncounter;
+    ComboBox *searcherSeason;
+    ComboBoxProxy *searcherLocation;
```
Admiral-Fish: Why are we declaring variables for something in the Ui namespace?

Form/Gen5/Pickup.cpp
```diff
@@ -0,0 +1,704 @@
...
+    ui->filterGenerator->disableControls(Controls::Height | Controls::Weight);
+
+    generatorIVAdvances = ui->textBoxGeneratorIVAdvances;
+    generatorEncounter = ui->comboBoxGeneratorEncounter;
+    generatorSeason = ui->comboBoxGeneratorSeason;
+    generatorLocation = ui->comboBoxGeneratorLocation;
+    generatorPokemon = ui->comboBoxGeneratorPokemon;
+    searcherEncounter = ui->comboBoxSearcherEncounter;
+    searcherSeason = ui->comboBoxSearcherSeason;
+    searcherLocation = ui->comboBoxSearcherLocation;
```
Admiral-Fish: What does this accomplish?

Form/Gen5/Pickup.cpp
```diff
@@ -0,0 +1,704 @@
...
+    ui->gridLayout->setVerticalSpacing(10);
+    ui->gridLayout_3->setColumnStretch(1, 0);
+    ui->gridLayout_3->setColumnStretch(3, 0);
+    ui->gridLayout_3->setAlignment(Qt::AlignTop);
+    ui->gridLayout_19->setColumnStretch(0, 1);
+    ui->gridLayout_19->setColumnStretch(1, 0);
+    ui->gridLayout_19->setColumnStretch(2, 3);
+    ui->gridLayout_24->setColumnStretch(0, 1);
+    ui->gridLayout_24->setColumnStretch(1, 0);
+    ui->gridLayout_24->setColumnStretch(2, 0);
```
Admiral-Fish: I have never put stuff like this in anything other than .ui file. Is any of it necessary?

Form/Gen5/Pickup.cpp
```diff
@@ -0,0 +1,704 @@
...
+    generatorModel = new PickupGeneratorModel5(ui->tableViewGenerator);
+    ui->tableViewGenerator->setModel(generatorModel);
+    connect(ui->filterGenerator, &Filter::showStatsChanged, generatorModel, &PickupGeneratorModel5::setShowStats);
+
+    searcherModel = new PickupSearcherModel5(ui->tableViewSearcher);
+    proxyModel = new SortFilterProxyModel(ui->tableViewSearcher, searcherModel);
+    ui->tableViewSearcher->setModel(proxyModel);
+
+    ui->textBoxGeneratorSeed->setValues(InputType::Seed64Bit);
+    generatorIVAdvances->setValues(InputType::Advance32Bit);
```
Admiral-Fish: Why isn't this just using the element from the Ui?

Form/Gen5/Pickup.cpp
```diff
@@ -0,0 +1,704 @@
...
+    generatorLocation = ui->comboBoxGeneratorLocation;
+    generatorPokemon = ui->comboBoxGeneratorPokemon;
+    searcherEncounter = ui->comboBoxSearcherEncounter;
+    searcherSeason = ui->comboBoxSearcherSeason;
+    searcherLocation = ui->comboBoxSearcherLocation;
+
+    generatorEncounter->setup({ toInt(Encounter::Grass), toInt(Encounter::GrassDark), toInt(Encounter::Surfing) });
+    searcherEncounter->setup({ toInt(Encounter::Grass), toInt(Encounter::GrassDark), toInt(Encounter::Surfing) });
+    generatorLocation->enableAutoComplete();
+    searcherLocation->enableAutoComplete();
```
Admiral-Fish: Why isn't this just using the element from the Ui?

Form/Gen5/Pickup.cpp
```diff
@@ -0,0 +1,704 @@
...
+    });
+
+    updateSearcherItemLists();
+
+    connect(generatorEncounter, &QComboBox::currentIndexChanged, this, &Pickup::generatorEncounterIndexChanged);
+    connect(generatorSeason, &QComboBox::currentIndexChanged, this, &Pickup::generatorSeasonIndexChanged);
+    connect(generatorLocation, &QComboBox::currentIndexChanged, this, &Pickup::generatorLocationIndexChanged);
+    connect(generatorPokemon, &QComboBox::currentIndexChanged, this, &Pickup::generatorPokemonIndexChanged);
+    connect(searcherEncounter, &QComboBox::currentIndexChanged, this, &Pickup::searcherEncounterIndexChanged);
+    connect(searcherSeason, &QComboBox::currentIndexChanged, this, &Pickup::searcherSeasonIndexChanged);
```
Admiral-Fish: Why isn't this just using the element from the Ui?

NickPlayeZ:
> Why does this basically reimplement Wild RNG? It just feels wrong and I don't like that, why aren't we just displaying only the items.

Whenever a user misses their target, they need to be able to tell what RNG advance they hit even if no pickup pokemon is holding an item.

Its also very useful to see what pokemon / gender / level your target is supposed to have so you can immediately tell whether you hit your target advance once the battle starts without needing to finish the battle and check for pickup items after

Core/Gen5/Generators/PickupGenerator.hpp
NickPlayeZ: 2 of these are leftovers from an older, really messy ui implementation; will remove them

Form/Gen5/Pickup.hpp
NickPlayeZ: more leftovers of the messy ui

Form/Gen5/Pickup.cpp
NickPlayeZ: nothing anymore; more leftovers

Form/Gen5/Pickup.cpp
NickPlayeZ: Moved it to the ui file; also just leftover mess lol

Form/Gen5/Pickup.cpp
NickPlayeZ: leftovers

Form/Gen5/Pickup.cpp
NickPlayeZ: leftovers

Form/Gen5/Pickup.cpp
NickPlayeZ: leftovers

Admiral-Fish:
Would it make more sense to somehow incorporate this into just the normal Wild window? Or do you think it should be separate?

NickPlayeZ:
> Would it make more sense to somehow incorporate this into just the normal Wild window? Or do you think it should be separate?

I think it gotta be separate; there's not really any space for the pickup pokemon and item ui in the settings and filters otherwise

PR #595, opened 2026-07-17
[Gen 5] Pass Power changes

NickPlayeZ:
This PR adds the following pass power related changes:

-IV advance 0-1 cannot show pass power results, as activating the C-Gear causes 2 IV advances and is required for activating pass powers

-PIDrng advances 0-3 cannot show pass power results, as activating the C-gear causes 1 PIDrng advance and then during the time it takes to activate the power, at least 5 more PIDrng advances happened during my tests (did 5 emu pausebuffering tests, 4 of which had 5 advances, 1 of which had 6 - just to be sure, since its inconsistent, I made it 0-3 instead of 0-4)

-added the proper arrow symbols for english pass power levels (and the proper german + symbols for german pass power levels).


Admiral-Fish:
I have similar comments about multi search as I did the other two PRs. I also feel like this would really only be useful for Wild where it can impact the encounter slots. Pretty sure it is useless for Static. I question how useful it would be in grotto.

Also I wouldn't be willing to accept the multi search as is since it is N times slower calling generate() each time for provided powers.

I could probably take some of the other changes if you separated them out. Friendly reminder to keep unrelated tweaks from entirely new functionality.

NickPlayeZ:
> I have similar comments about multi search as I did the other two PRs. I also feel like this would really only be useful for Wild where it can impact the encounter slots. Pretty sure it is useless for Static. I question how useful it would be in grotto.
> 
> Also I wouldn't be willing to accept the multi search as is since it is N times slower calling generate() each time for provided powers.
> 
> I could probably take some of the other changes if you separated them out. Friendly reminder to keep unrelated tweaks from entirely new functionality.

For Static it lets users search for no pass power and lucky power 3 at the same time.
For Grotto, the user can also search for the lower level grotto powers at the same time in case their specific target is available with a lower pass power as well, not just level 3.
Same goes for wild and there it'll be even more useful for step encounter stuff if a specific target requires both lucky power and encounter power at the same time.

I'll isolate the more minor changes into this PR and then make a new PR where I'll make the multi pass power search behave more closely to the way you wanted me to change the multi lead search

Test/Gen5/WildGenerator5Test.cpp
```diff
@@ -72,6 +72,10 @@ void WildGenerator5Test::generate()
     QFETCH(std::string, results);
 
     json j = json::parse(results);
+    if (luckyPower != PassPower::None)
+    {
+        j = json::array();
+    }
```
Admiral-Fish: This is just disabling/modifying test cases/data out of the input data? Seems like a bad idea.

Do existing test need to be modifying or new ones created?

Test/Gen5/StaticGenerator5Test.cpp
```diff
@@ -123,6 +123,10 @@ void StaticGenerator5Test::generateWild()
     QFETCH(std::string, results);
 
     json j = json::parse(results);
+    if (luckyPower != PassPower::None)
+    {
+        j = json::array();
+    }
```
Admiral-Fish: This is just disabling/modifying test cases/data out of the input data? Seems like a bad idea.

Do existing test need to be modifying or new ones created?

Test/Gen5/HiddenGrottoGeneratorTest.cpp
```diff
@@ -147,6 +148,19 @@ void HiddenGrottoGeneratorTest::slot()
...
+        u32 minAdvance = Utilities5::initialAdvances(seed, profile) + 4;
+        for (const auto &state : j)
+        {
+            if (state["advances"].get<u32>() >= minAdvance)
+            {
+                filtered.emplace_back(state);
+            }
+        }
+        j = filtered;
+    }
```
Admiral-Fish: This is just disabling/modifying test cases/data out of the input data? Seems like a bad idea.

Do existing test need to be modifying or new ones created?

Core/Gen5/Generators/WildGenerator5.cpp
```diff
@@ -127,6 +127,20 @@ std::vector<WildState5> WildGenerator5::generate(u64 seed, u32 initialAdvances,
...
+        for (const auto &iv : ivs)
+        {
+            if (iv.first >= 2)
+            {
+                powerIVs.emplace_back(iv);
+            }
+        }
+        validIVs = &powerIVs;
+    }
+
```
Admiral-Fish, line 138: This can probably be done like this instead.

if (luckyPower != PassPower::None) {
    std::erase_if(ivs, [](const auto &iv) { return iv.first < 2; });
}

Core/Gen5/Generators/WildGenerator5.cpp
```diff
@@ -109,7 +109,7 @@ std::vector<WildState5> WildGenerator5::generate(u64 seed, u32 initialAdvances,
     {
         std::array<u8, 6> iv;
         std::ranges::generate(iv, [&rngList] { return rngList.next(); });
-        if (filter.compareIV(iv))
+        if (filter.compareIV(iv) && (luckyPower == PassPower::None || initialAdvances + cnt >= 2))
```
Admiral-Fish: We are double filtering. Not sure we need this one as well.

Core/Gen5/Generators/StaticGenerator5.cpp
```diff
@@ -166,6 +166,20 @@ std::vector<State5> StaticGenerator5::generateNonWild(u64 seed, const std::vecto
...
+        for (const auto &iv : ivs)
+        {
+            if (iv.first >= 2)
+            {
+                powerIVs.emplace_back(iv);
+            }
+        }
+        validIVs = &powerIVs;
+    }
+
```
Admiral-Fish, line 177: This can probably be done like this instead.

if (luckyPower != PassPower::None) {
std::erase_if(ivs, [](const auto &iv) { return iv.first < 2; });
}

Core/Gen5/Generators/StaticGenerator5.cpp
```diff
@@ -87,7 +87,7 @@ std::vector<State5> StaticGenerator5::generate(u64 seed, u32 initialAdvances, u3
             iv[5] = rngList.next();
         }
 
-        if (filter.compareIV(iv))
+        if (filter.compareIV(iv) && (luckyPower == PassPower::None || !staticTemplate.getWild() || initialAdvances + cnt >= 2))
```
Admiral-Fish: We are double filtering. Not sure we need this one as well.

Core/Gen5/Generators/StaticGenerator5.cpp
```diff
@@ -243,7 +257,12 @@ std::vector<State5> StaticGenerator5::generateWild(u64 seed, const std::vector<s
         }
 
         u32 prng = rng.nextUInt();
-        for (const auto &iv : ivs)
+        if (luckyPower != PassPower::None && initialAdvances + cnt < 4)
+        {
+            continue;
+        }
```
Admiral-Fish: What is this check for and why is it so late down the execution flow?

Core/Gen5/Generators/HiddenGrottoGenerator.cpp
```diff
@@ -82,6 +82,11 @@ std::vector<HiddenGrottoState> HiddenGrottoSlotGenerator::generate(u64 seed) con
     {
         BWRNG go(rng, jump);
         u32 prng = rng.nextUInt();
+        if (grottoPower != PassPower::None && initialAdvances + cnt < 4)
+        {
+            continue;
+        }
+
```
Admiral-Fish: After reviewing the other places you have this check I am spotting a similar pattern. It probably makes more sense to just modify what `u32 cnt = 0;` starts at in the for loop instead of checking within the loop. It will be much better for performance as we will remove an necessary conditional statement running each iteration when we don't really need to.

Might require advancing the `rng` by whatever we have to start `cnt` at.

Admiral-Fish (changes requested):

Test/Gen5/WildGenerator5Test.cpp
NickPlayeZ: I forgot to upload all the test fixes; ups

Core/Gen5/Generators/WildGenerator5.cpp
NickPlayeZ: it wouldnt compile that way, so instead I did:

std::vector<std::pair<u32, std::array<u8, 6>>> powerIVs;
const auto *validIVs = &ivs;

if (luckyPower != PassPower::None)
{
    powerIVs = ivs;
    std::erase_if(powerIVs, [](const auto &iv) { return iv.first < 2; });
    validIVs = &powerIVs;
}

Core/Gen5/Generators/WildGenerator5.cpp
NickPlayeZ: we indeed dont need the double; gonna fix that

Core/Gen5/Generators/StaticGenerator5.cpp
NickPlayeZ: its for the PIDrng advances of 4 or higher being needed for pass powers to be valid; but yeah too far down. Will fix

Core/Gen5/Generators/HiddenGrottoGenerator.cpp
NickPlayeZ: changed as requested

PR #598, opened 2026-07-21
[Gen 5] Allow multi item search for Hidden Grottoes

NickPlayeZ:
Makes makes genderless pokemon not display male/female unnecessarily, simplifies the pokemon tabs and lets users search for multiple of the same item to e.g. make PP Max farming easy and quick

grafik

grafik

### DIFF
--- a/Core/Gen5/Generators/HiddenGrottoGenerator.cpp
+++ b/Core/Gen5/Generators/HiddenGrottoGenerator.cpp
@@ -77,7 +77,7 @@ std::vector<HiddenGrottoState> HiddenGrottoSlotGenerator::generate(u64 seed) con
             u8 slot = encounterTable[go.nextUInt(100)];
             if (slot < 3) // Pokemon
             {
-                const auto &pokemon = encounterArea.getPokemon(group, slot);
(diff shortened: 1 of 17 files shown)


Admiral-Fish:
What was broken about grotto power? I can't tell. The fixes should be separate from functionality for review purposes, especially since I'm not sure I like the concept of searching in generator as I mentioned on the other PR

NickPlayeZ:
> What was broken about grotto power? I can't tell. The fixes should be separate from functionality for review purposes, especially since I'm not sure I like the concept of searching in generator as I mentioned on the other PR

Previously, grotto power only increased the advances that stuff appears on, but it didnt take into account the shift in odds for pokemon/item/hidden item which goes from 20/40/40 to 49.8/25.6/25.6


If you have a better idea of how to implement multi item search be my guest, but this is an extremely useful feature that will allow people to pp max farm super quickly

Admiral-Fish:
My point is still that those fixes should be separate from additional functionality. Any issues I have with the new functionality will delay potentially legitimate fixes.

I wonder if this multi item deserves like a separate class. It just feels like it is being injected into afterwards as a sloppy afterthought.

NickPlayeZ:
Separated the two now

PR #599, opened 2026-07-22
[Gen 4] Add hgss rock smash items & modifiers to generator only

NickPlayeZ:
I felt it was only useful to add to the generator, as the items arent really particularly rare.

Similar case to wild held items where its just not necessary to search for an item specifically and you can very easily just find what you need.

But with this implementation now users can RNG for any random seed and purposely target any specific items they may want to get from rock smash


Admiral-Fish:
This almost feels like something to me that would be better in a separate utility tool

NickPlayeZ:
> This almost feels like something to me that would be better in a separate utility tool

But wouldn't it be nice to have as many rng manip features within pokefinder as possible (that are of some use)?

Admiral-Fish:
I mean as a seperate window within pokefinder. I don't see the value of adding it straight to the wild window

NickPlayeZ:
> I mean as a seperate window within pokefinder. I don't see the value of adding it straight to the wild window

I figured since it's such a small feature, adding it to wild rng would be best. Especially since those already have the regular rock smash functionality.
And it may help users figure out what they hit instead of their target when going for a pokemon if the items of the relevant advances are shown.

But I can make it its own window if needed

PR #600, opened 2026-07-27
[Gen 5] Fix Grotto Power

NickPlayeZ:
Was broken previously not accounting for the effect Grotto Power had on what is generated


Admiral-Fish:
Closing in favor of e3056a4e3f4a916447d799d3b3cb9f827a030e6a

PR #601, opened 2026-07-29
[Gen 5] Fix adjacent seeds functionality + add IV filters and sorting

NickPlayeZ:
Adds +/-1 timer0 so adjacent seeds can fulfill its primary purpose properly letting users see when they missed their timer0

Also adds IV filters and IV sorting


Form/Gen5/Tools/AdjacentSeeds.cpp
```diff
@@ -119,6 +126,23 @@ void AdjacentSeeds::updateProfiles()
...
+        if (minIVWidgets[i]->value() > maxIVWidgets[i]->value())
+        {
+            QMessageBox msg(QMessageBox::Warning, tr("Invalid IV settings"), tr("%1 minimum is greater than maximum").arg(stats[i]));
+            msg.exec();
+            return;
+        }
+    }
+
+    ui->tableView->horizontalHeader()->setSortIndicator(-1, Qt::AscendingOrder);
+    proxyModel->sort(-1);
```
Admiral-Fish: What does this accomplish?

NickPlayeZ: It resets any sorting the user applied to the columns when clicking the "generate" button. Otherwise there's no option to reset sorting swiftly

Core/Gen5/Tools/AdjacentSeedsCalculator.cpp
```diff
@@ -57,8 +57,8 @@ namespace AdjacentSeedsCalculator
     std::vector<AdjacentSeedsState> generate(u32 minIVAdvance, u32 maxIVAdvance, int seconds, bool roamer, Buttons buttons,
                                              const DateTime &dateTime, const Profile5 &profile)
     {
-        u32 timer0Min = profile.getTimer0Min();
-        u32 timer0Max = profile.getTimer0Max();
```
Admiral-Fish: I'm still not sure I agree with this.  If your profile doesn't have the correct timer0 range then it should be adjusted, not just some arbitrary +/-1 range

NickPlayeZ: this is how it always worked in rng reporter and its really useful for BW1 players especially.
Lets say a BW1 players primary timer0 "happens" about 70% of the time and the other timer0 about 30% of the time.
In such a case, the player would almost always want to search for seeds with that more common timer0 only, not the less common one.
Now if we didnt have this +/-1 timer0 "extension", he would have to either make a copy of his profile for adjacent seeds and change the timer0 range, which can get crowded in the profile manager if the player rngs on multiple games, or he has to always manually change the timer0 range just to use adjacent seeds proper.
If we simply account for +/-1 timer0, the player doesnt have to do anything extra and itll just work.

Another case could be when the player does not know their less common timer0. Perhaps after 10 profile calibrations he only ever got C7B. Now the other timer0 could be C7A or C7C, but if he's brand new to rng he may not even know that.
If we just default to +/-1 timer0, adjacent seeds will already work perfectly and display all the possible seeds the player can hit assuming proper execution, even if he doesnt know the other timer0.

All in all, its an important quality of life change

Admiral-Fish: I believe the original argument was just RNG reporter which I don't really care about. This explanation makes more sense.

Admiral-Fish:
This should be good to merge now. I moved the filtering into the backend code and swapping to using the IVFilter class I split out to promote code reuse.

Let me know if you think anything else should change before this merges

Form/Gen5/Tools/AdjacentSeeds.cpp
Admiral-Fish: I removed the proxyModel sort as the model is immediately cleared. Didn't seem to impact the intention of resetting the sort order.

NickPlayeZ:
> This should be good to merge now. I moved the filtering into the backend code and swapping to using the IVFilter class I split out to promote code reuse.
> 
> Let me know if you think anything else should change before this merges

Why did you remove the +/-1 timer0 again and why remove the reset IVs button?
Theyre both important for the functionality of this so I'm gonna add them back.
The timer0 i explained last time.
And for the reset IVs button, its very annoying to try and get an overview over all adjacent seeds again after inputting IV ranges without it cause you'd have to manually reset each range to 0-31

Admiral-Fish:
Woops, I deleted the timer0 thing on accident. As for reset IVs it's redundant as the IV filter has controls for that by clicking on the labels for each IV

NickPlayeZ:
> Woops, I deleted the timer0 thing on accident. As for reset IVs it's redundant as the IV filter has controls for that by clicking on the labels for each IV

Hmm a bit less intuitive with the IVs, but ig its fair enough being in line with the searcher and generator tabs functionality

Admiral-Fish (approved):

PR #605, opened 2026-08-26
Update Japanese translation

norimaki7:
Completes the Japanese translation.

- UI strings: 2352/2352
- Location names for all 9 games
- Move names, form names, game names


Admiral-Fish:
I see a co-author by Claude, what was AI used for here? Did you do all of the translations yourself?

norimaki7:
It did the base translation, and I reworded it into the terms the Japanese
RNG community actually uses. Not every line was typed by me, but the
terminology is mine and I reviewed the whole thing in a local build.

Admiral-Fish:
Do you speak Japanese? Just wondering how much of these I need to try to verify.

norimaki7:
Yes, Japanese is my first language.

The names — Pokémon, moves, abilities, natures, types, items — came straight
from PokeAPI's Japanese names, so those are mechanically correct and don't
need checking. The UI wording is where my judgement comes in, so that's the
part to look at if you want to spot-check anything.

Admiral-Fish:
Just wanted to make sure these translations were at least coming from someone that speaks the language. I can't easily verify since I only speak English.

So if I am following correctly, you got the easy things from PokeAPI and the UI translations you used AI as a base then went through and cleaned up with your understanding of the language and community terms?

PR #606, opened 2026-08-27
Fixed Honey Tree Generator and Searcher

domx9200:
Because of the way the getCheckedArray function filled the array, it filled any extra slots with "true" which meant the honey tree array, which had 6 entries, and also required 1 encounter to be selected only, would never work. as there would be a minimum of 6, when nothing was selected, or 7 when the expected number was selected. This just simply added a boolean to allow the backfill to be false instead of true optionally, as to prevent breaking changes.

### DIFF
--- a/Form/Controls/CheckList.hpp
+++ b/Form/Controls/CheckList.hpp
@@ -94,14 +94,16 @@ public:
      *
      * @tparam size Size of the array
      *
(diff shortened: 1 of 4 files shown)


domx9200:
fixes #607

Admiral-Fish:
Fixed via 666d5bd8e17843b3765eefcde98a3de4fd7a2e97

PR #610, opened 2026-09-14
[Gen 5] Add egg spawn prediction

NickPlayeZ:
This adds an egg column to the gen 5 eggs generator for bw2 profiles only.
If there's a "Yes", that advance will yield an egg to spawn.

This is useful for BW2 only, as in BW2 the egg rng is done in 2 steps where first you have to spawn the egg at the daycare man and it'll already have its IVs, nature and ability set.

To spawn an egg, you gotta open up any sub-menu like your party (this resets the step counter), take 16 steps, advance until you hear the chatot of 1 before a Yes egg advance and then take the 17th step to spawn it (these games check for whether an egg spawns every 17th step).

Had to change the ui slightly to make space for the parent species and a new original trainer same/different selector, as those impact how many advances show a Yes in the egg column. For this, I also added the oval charm to bw2 profiles (and made the shiny charm need the oval charm, since you cant get the shiny charm without having first gotten the oval charm ingame).


Admiral-Fish:
Honestly this seems like overkill to me. Refer to how I handle this in Gen 3: there is just a dropdown where you enter the compatibility (contains the 20/50/70 values and text that is displayed in game when you talk to the daycare person).

Would remove most of the stuff you had to add (parent species, original/different trainer probably everything in EggSettings that would spiral out to other places). Would only keep the oval charm which would modify the compatibility.

Also if there a compelling reason to not add this for BW as well.

NickPlayeZ:
> Honestly this seems like overkill to me. Refer to how I handle this in Gen 3: there is just a dropdown where you enter the compatibility (contains the 20/50/70 values and text that is displayed in game when you talk to the daycare person).
> 
> Would remove most of the stuff you had to add (parent species, original/different trainer probably everything in EggSettings that would spiral out to other places). Would only keep the oval charm which would modify the compatibility.
> 
> Also if there a compelling reason to not add this for BW as well.

Ig you could do that way but I feel this way even without an rng tutorial, people will be able to use, whilst with just the values it'll be a lot less clear.

Its just not useful in bw1; you just spawn the egg there first, save the game and then do the rng. Ig people could want to rng multiple eggs in a row for some reason, though there getting specific IVs or shinies will be difficult. Not opposed to adding it to bw1, just don't see any use case

Admiral-Fish:
I think having to select 1 thing that is said in game if you talk to one person instead of 3 things is less to manage.

It also adds what I believe to be extra code complexity for no real gain. And it also only handles gen 5. What if we want to add this to other gens. It also doesn't match gen 3 so now we have a break in consistency.

Besides the egg generator infustructure was setup with the gen 3 way in mind.

NickPlayeZ:
> I think having to select 1 thing that is said in game if you talk to one person instead of 3 things is less to manage.
> 
> It also adds what I believe to be extra code complexity for no real gain. And it also only handles gen 5. What if we want to add this to other gens. It also doesn't match gen 3 so now we have a break in consistency.
> 
> Besides the egg generator infustructure was setup with the gen 3 way in mind.

I don't feel too strongly about it one way or another; would prefer the way I did it but if you wanna do it differently thats no problem here either

NickPlayeZ:
changes made; on second thought, this might be better

Admiral-Fish (approved):

